Issue #15, opened 2020-12-02
Update Dashboard for December Update


philipstanislaus:
- [x] header: add explainer drop [ui]
- [x] header: add total assets locked (all finances with status open or ongoing) [ui]
- [x] header: add total value (reserve + NAV) [ui]
- [ ] header: add rewards [ui]
- [x] table: update row design [ui]
- [x] table: icons for every pool [ui]
- [x] table: pool value [ui]
- [x] table: label upcoming and oversubscribed [ui]
- [ ] table: remove legacy pools
- [x] add projected 14 day drop yield to upcoming pools [pool config, ui]
- [x] add projected 14 day tin yield to upcoming pools [pool config, ui]

PR #19, opened 2020-12-04
Dashboard December update

philipstanislaus:


tinlake-ui/README.md
```diff
@@ -17,6 +17,9 @@ add the following env variables
 
 ### Run
 
+To build tinlake.js and start the NextJs server and serve the Netlify lambdas locally
+
 ```shell
-yarn start # to start the NextJs server and serve the Netlify lambdas locally
+../tinlake.js; yarn build; cd ../tinlake-ui
```
offerijns, line 23: Yeah I actually somehow integrated this already (see the `build:deps` script in `tinlake-ui/package.json`). It should be possible to automatically build tinlake.js while building tinlake-ui, but it needs some workarounds. The "native" feature for this is TS project references, but Next.js doesn't support that. This might be the workaround we need: https://josephluck.co.uk/blog/next-typescript-monorepo. Anyhow, I'll add this as a ticket on the board.

tinlake-ui/components/PoolList/index.tsx
```diff
@@ -20,84 +19,173 @@ class PoolList extends React.Component<Props> {
...
+          </PoolRow>
+        ))}
       </Box>
     )
   }
 }
 
 export default PoolList
+
+const Header = styled.div`
```
offerijns, line 104: Maybe these styles can be moved to a separate file, `styles.ts`? Might be a bit cleaner.

tinlake-ui/components/PoolsMetrics/index.tsx
```diff
@@ -13,29 +13,60 @@ class PoolsMetrics extends React.Component<Props> {
...
+          <Label>Total value locked</Label>
+        </Box>
+      </>
     )
   }
 }
 
 export default PoolsMetrics
+
+const Cont = styled.div`
```
offerijns, line 40: Same as the other comment, I think it would be better to separate this.

tinlake-ui/components/TinlakeExplainer/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+          of these real-world assets.
+        </Text>
+      )}
+    </Container>
+  )
+}
+
+export default TinlakeExplainer
+
+const Container = styled.div`
```
offerijns, line 37: Same as the other comment, I think it would be better to separate this.

offerijns (changes requested):
The code changes look good overall! Just left a few semantic comments. But I think the UI design is not finished yet right? I checked it out locally and it looks like there's some work remaining.

tinlake-ui/README.md
philipstanislaus: Cool, removed from README and added your `build:deps` script to `yarn start`: a2e0ed00f75a7ff0889c4c14dadc5ab44ce10f85

tinlake-ui/components/PoolList/index.tsx
philipstanislaus: Done

philipstanislaus:
Thanks for the review @offerijns , implemented all your fixes. 

A few things are still open: 

1. Updating the logic for oversubscription. Right now, it's just showing as oversubscribed if reserve > maxReserve.
2. Adding rewards, which we will do as a separate iteration, not as part of this PR.
3. Removing legacy pools. They are still in the current dashboard for discussing what exactly we want to show.
4. Removing the DROP APR column. As 3, I left it for now for discussing what we want to show.

This issue closes #15.

offerijns (approved):
Looks great! I have some small UI changes I'd like to make, but let's just get this merged first and that can be done later then :) I would already remove the legacy pools now though, since this cleans up the UI a lot and makes it easier to view this new dashboard design as it's intended.

PR #28, opened 2020-12-07
added ipfs hash fetching method to tinlake.js

### LINKED ISSUE #8: Load pools in UI from IPFS at build time

charleenfei:
closes #8


charleenfei:
closes #8

tinlake-ui/next.config.js
```diff
@@ -16,47 +16,5 @@ module.exports = {
   },
   experimental: {
     exportTrailingSlash: false,
-  },
```
charleenfei, line 19: i had to remove this redirect section due to the new async way we have of fetching the pool config

charleenfei:
closes #8

offerijns (changes requested):
Code looks great, just a few comments! As mentioned on Slack, we do need to first push the latest mainnet pools to the registry, and then double check that everything works correctly, before we can merge this.

philipstanislaus (commented):
Nice, thanks @charleenfei ! 🎉 

Thinking about this architecture some more, I wonder whether it makes sense to dynamically load data from IPFS on the client side.

If I understand this correctly, the pool pages can't dynamically be updated anyways (through a new entry in the on-chain pool registry and new data on IPFS), since we statically create the paths at build time. 

If we wanted dynamic pools, we'd have to remove static paths or change the static path logic to use dynamic fallback (see [Next.js docs for more](https://nextjs.org/docs/basic-features/data-fetching#the-fallback-key-required)). 

What has been implemented here should impact the page's speed quite a bit, since there is 1 request to an Infura node, and afterwards 1 request to an IPFS node. Both of those are done on every page load from the client side. 

We could mitigate that performance hit by using server side rendering and cache the responses to the Ethereum/IPFS requests to happen every 15 seconds in the background.

Alternatively, we could use your logic in the static path hooks only to stay completely static. This has the upside of being very simple, fast, but requires a rebuild/export of the app whenever we add/remove pools.

Can we quickly discuss those trade-offs? Thanks!

tinlake-ui/components/PoolLink/index.tsx
```diff
@@ -1,75 +1,59 @@
...
-function getAs(rootOrSlug: string | string[], as: string | null | undefined | UrlObject): string {
-  const pool = config.pools.find(
-    (p) => (rootOrSlug as string).toLowerCase() === p.addresses.ROOT_CONTRACT.toLowerCase()
-  )
-  if (pool) {
-    return `/pool/${rootOrSlug}/${pool.metadata.slug}${as}`
+function getAs(
+  root: string | string[] | undefined,
+  slug: string | string[] | undefined,
+  as: string | null | undefined | UrlObject
```
philipstanislaus, line 53: Again, I think we should change them to: 

```suggestion
  root: string | undefined,
  slug: string | undefined,
  as: string | null | undefined
```

tinlake-ui/components/PoolLink/index.tsx
```diff
@@ -1,75 +1,59 @@
...
   )
 }
 
-function getHref(rootOrSlug: string | string[], href: string | null | undefined | UrlObject): string {
-  console.log("GETHREF", config.pools, config.upcomingPools)
-  const pool = config.pools.find(
-    (p) => (rootOrSlug as string).toLowerCase() === p.addresses.ROOT_CONTRACT.toLowerCase()
-  )
-  if (pool) {
+function getHref(root: string | string[] | undefined, href: string | null | undefined | UrlObject): string {
```
philipstanislaus, line 43: These types are not correct anymore, right? I guess the following would be right: 

```suggestion
function getHref(root: string | undefined, href: string | null | undefined): string {
```

tinlake-ui/components/PoolLink/index.tsx
```diff
@@ -1,75 +1,59 @@
...
+  if (root) {
```
philipstanislaus, line 44: This needs to be a check for `slug`, right?

```suggestion
  if (slug) {
```

tinlake-ui/components/PoolLink/index.tsx
```diff
@@ -1,75 +1,59 @@
...
-    (p) => (rootOrSlug as string).toLowerCase() === p.addresses.ROOT_CONTRACT.toLowerCase()
-  )
-  if (pool) {
-    return `/pool/${rootOrSlug}/${pool.metadata.slug}${as}`
+function getAs(
+  root: string | string[] | undefined,
+  slug: string | string[] | undefined,
+  as: string | null | undefined | UrlObject
+): string {
+  if (root) {
```
philipstanislaus, line 55: This should test for `slug` again, right? 

```suggestion
  if (slug) {
```

tinlake-ui/components/ValueNFT/index.tsx
```diff
@@ -24,6 +24,7 @@ interface Props extends TransactionProps {
 const DAYS = 24 * 60 * 60 * 1000
 
 const ValueNFT: React.FC<Props> = (props: Props) => {
+  console.log(props.tokenId, 'TOKEN ID VALUE')
```
philipstanislaus, line 27: Can we remove this?

philipstanislaus (approved):
Nice, just a few minor comments 🎉

tinlake-ui/components/PoolLink/index.tsx
charleenfei: the types are correct, the typing for router.query.root must also include string[]

tinlake-ui/components/PoolLink/index.tsx
charleenfei: please see comment above

offerijns (approved):
LGTM!

PR #99, opened 2021-01-11
feat: finalize onboarding MVP, add portfolio page

offerijns:


onboard-api/src/controllers/address.controller.ts
```diff
@@ -39,8 +43,8 @@ export class AddressController {
     if (kyc) {
       let status: KycStatusLabel = kyc.status
 
-      if (!kyc.verifiedAt) {
-        const investor = await this.securitizeService.getInvestor(kyc.digest)
+      if (kyc.status !== 'verified' || (kyc.usaTaxResident && !kyc.accredited)) {
```
charleenfei, line 46: we could set 'verified' as a const here somewhere in a const list of potential statuses, in addition to the typing you have created. this would also us to make changes more uniformly in the future if we need to update statuses.

onboard-api/src/controllers/agreement.controller.ts
```diff
@@ -54,6 +59,17 @@ export class AgreementController {
   async postDocusignConnect(@Body() body): Promise<string> {
     console.log('Received Docusign Connect message')
     console.log(body)
+
+    const content = JSON.parse(body)
+    const investor = content.recipients.signers.find((signer: any) => signer.roleName === InvestorRoleName)
+    const issuer = content.recipients.signers.find((signer: any) => signer.roleName === IssuerRoleName)
+
+    const status = {
+      signed: investor?.status === 'completed',
```
charleenfei, line 68: see status comment above

onboard-api/src/controllers/agreement.controller.ts
```diff
@@ -54,6 +59,17 @@ export class AgreementController {
...
+      counterSigned: issuer?.status === 'completed',
+    }
+    console.log({ status })
```
charleenfei, line 71: console is intentional?

onboard-api/src/controllers/kyc.controller.ts
```diff
@@ -35,23 +38,41 @@ export class KycController {
...
+    // Update KYC and user records in our database
     const kyc = await this.kycRepo.upsertSecuritize(address.userId, kycInfo.providerAccountId, kycInfo.digest)
     if (!kyc) throw new BadRequestException('Failed to create KYC entity')
 
     await this.userRepo.update(address.userId, investor.email, investor.fullName, investor.details.address.countryCode)
 
-    await this.agreementRepo.createAgreementsForPool(params.poolId, address.userId, investor.email)
+    this.kycRepo.setStatus(
+      'securitize',
+      kyc.providerAccountId,
```
charleenfei, line 58: same as above

onboard-api/src/controllers/address.controller.ts
```diff
@@ -53,12 +57,23 @@ export class AddressController {
...
-          this.kycRepo.setStatus('securitize', kyc.providerAccountId, investor.verificationStatus as KycStatusLabel)
-          status = investor.verificationStatus as KycStatusLabel
+          this.kycRepo.setStatus(
+            'securitize',
+            kyc.providerAccountId,
+            investor.verificationStatus === 'manual-review' ? 'processing' : investor.verificationStatus,
+            investor.domainInvestorDetails.isUsaTaxResident,
+            investor.domainInvestorDetails.isAccredited
+          )
+          status = investor.verificationStatus === 'manual-review' ? 'processing' : investor.verificationStatus
```
charleenfei, line 67: same for these verification statuses

charleenfei (approved):
small changes but otherwise lgtm!

onboard-api/src/controllers/address.controller.ts
offerijns: I think this adds a lot of overhead (you'd need many const, `KycVerified`, `KycProcessing`, etc...), while not bringing much benefit. It's already typed through `KycStatusLabel`, so let's say we rename 'verified' to 'done', then typescript would already throw an error here, something like `'verified' is not a possible status of enum type KycStatusLabel`.

onboard-api/src/controllers/agreement.controller.ts
offerijns: Yeah, still working on this endpoint, so I'm logging the output for now.

PR #119, opened 2021-01-21
fix: agreement creation flow

offerijns:


onboard-api/src/controllers/address.controller.ts
```diff
@@ -66,27 +67,39 @@ export class AddressController {
...
-      )
-
-      const agreementLinks = agreements.map(
-        (agreement: Agreement): AgreementsStatus => {
+      // Filter profile agreements by country
+      const profileAgreements = pool.profile?.agreements
+        .filter((pa: ProfileAgreement) => {
+          return (
+            (user.countryCode === 'US' && pa.country === 'non-us') || (user.countryCode !== 'US' && pa.country === 'us')
+          )
```
philipstanislaus, line 75: Shall we throw an error here if these filters don't match?

onboard-api/src/controllers/address.controller.ts
```diff
@@ -66,27 +67,39 @@ export class AddressController {
...
+      // Retrieve
+      const agreementLinks = await this.agreementRepo.getStatusForProfileAgreements(
+        user.id,
+        params.poolId,
+        profileAgreements
       )
 
       const isWhitelisted = await this.investmentRepo.getWhitelistStatus(address.id, params.poolId)
 
+      // TODO: this should also filter by blockchain and network
```
philipstanislaus, line 95: Should we create a ticket for this? Could be a "good first issue" for one of our new hires.

onboard-api/src/services/pool.service.ts
```diff
@@ -51,7 +54,14 @@ export class PoolService {
     )
 
     this.pools = poolsWithProfiles
-    this.logger.log(`Loaded ${Object.keys(this.pools).length} pools with profiles from IPFS`)
+    const newPools = Object.values(poolsWithProfiles).filter((pool: Pool) => !prevPools.includes(pool))
+
+    if (newPools.length > 0) this.logger.log(`Loaded ${Object.keys(this.pools).length} pools with profiles from IPFS`)
```
philipstanislaus, line 59:
```suggestion
    if (newPools.length > 0) this.logger.log(`Loaded ${Object.keys(this.pools).length} new pools with profiles from IPFS`)
```

onboard-api/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Agreement, AgreementRepo } from '../repos/agreement.repo'
+import { UserRepo, UserWithKyc } from '../repos/user.repo'
+
+@Controller()
+export class UserController {
+  constructor(private readonly agreementRepo: AgreementRepo, private readonly userRepo: UserRepo) {}
+
+  @Get('users/:poolId')
+  async getUsers(@Param() params) {
+    return {}
```
philipstanislaus, line 12: Is a TODO missing here?

philipstanislaus (approved):
LGTM, just a few minor comments 👍

tinlake.js/bin/decode.js
```diff
@@ -0,0 +1,21 @@
...
+
+const run = async () => {
+  const decoded = contract.interface.decodeFunctionData('borrow', '0x0ecbcdab0000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000675e56da063e8940000')
+  decoded.forEach((value) => {
+    console.log(value.toString())
+  })
+
+}
+
+run()
```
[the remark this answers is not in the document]
offerijns: Why not? I think it's a useful debugging script, that I imagine I'll definitely also use in the future.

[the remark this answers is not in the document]
philipstanislaus: Okay :)

onboard-api/src/controllers/address.controller.ts
offerijns: That wouldn't be an error, this is just a filter to get the US agreement for US based investors and vice versa.

onboard-api/src/controllers/address.controller.ts
offerijns: I would say touching this codebase shouldn't be a first issue actually, it requires quite a lot of understanding of the onboarding process and how all the integrations fit together.

onboard-api/src/controllers/user.controller.ts
offerijns: Yes :)

PR #123, opened 2021-01-28
fix: show last 90 days of pool data on dashboard

offerijns:


tinlake-ui/services/apollo/index.ts
```diff
@@ -448,17 +448,14 @@ class Apollo {
...
-      }
-    })
+    const poolsDailyData = result.data.days
+      .map((item: any) => {
+        return {
+          day: Number(item.id),
+          poolValue: parseFloat(new BN(item.assetValue).add(new BN(item.reserve)).div(UintBase).toString()),
+        }
+      })
+      .sort((a: PoolsDailyData, b: PoolsDailyData) => a.day - b.day)
```
philipstanislaus, line 458: Why is sorting needed here? The query already sorts them, right?

philipstanislaus (approved):

PR #150, opened 2021-02-23
feat: pool overview page

offerijns:


tinlake-ui/pages/_app.tsx
```diff
@@ -35,6 +35,21 @@ class MyApp extends App {
...
+              _paq.push(['enableLinkTracking']);
+              (function() {
+                var u="https://centrifuge.matomo.cloud/";
+                _paq.push(['setTrackerUrl', u+'matomo.php']);
+                _paq.push(['setSiteId', '2']);
+                var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+                g.type='text/javascript'; g.async=true; g.src='//cdn.matomo.cloud/centrifuge.matomo.cloud/matomo.js'; s.parentNode.insertBefore(g,s);
+              })();`,
+            }}
+          />
```
philipstanislaus, line 52: Can we put the tracking/site ID into an env variable, and use a different value for dev/staging than production?

philipstanislaus (approved):
LGTM, just one comment :)

tinlake-ui/pages/_app.tsx
offerijns: Good idea 👍 Also helpful to not have it installed when developing locally.

PR #151, opened 2021-02-25
fix: allow claims to be collected whenever there are claimable reward…

philipstanislaus:
…s on the last link, improve debug box

### DIFF
--- a/tinlake-ui/containers/UserRewards/index.tsx
+++ b/tinlake-ui/containers/UserRewards/index.tsx
@@ -5,7 +5,6 @@ import { useRouter } from 'next/router'
 import * as React from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import styled from 'styled-components'
-import Alert from '../../components/Alert'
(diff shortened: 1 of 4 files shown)


tinlake-ui/containers/UserRewards/index.tsx
```diff
@@ -228,31 +238,69 @@ const UserRewards: React.FC<Props> = ({ tinlake }: Props) => {
 
 export default UserRewards
 
-const days = 24 * 60 * 60
+const day = 24 * 60 * 60
+const minNonZeroDays = 61
```
offerijns, line 242: I think the 60 days should be put in an env var or if possible, retrieved from the subgraph. Since it looks like it will change to 30 days soon.

offerijns (approved):

tinlake-ui/containers/UserRewards/index.tsx
philipstanislaus: Good point, I created #152 for that. Thanks!

PR #177, opened 2021-04-03
feat: implement multicall for portfolio page

### LINKED ISSUE #113: Use multicall to update portfolio prices and values in real time
The portfolio page displays balances and prices of the token holdings of a specific address. An example is https://tinlake.centrifuge.io/portfolio?address=0x683b73162ca9f344847c01d5040f07926b4d1378.

This data is retrieved from the subgraph, but the subgraph only updates the balances on a daily basis. See https://github.com/centrifuge/tinlake-subgraph/blob/main/src/domain/DailyPoolData.ts#L12

We want to fix this by using multicall to retrieve the real-time balance and price. An example usage of multicall in the tinlake UI codebase is here: https://github.com/centrifuge/apps/blob/main/tinlake-ui/ducks/pool.ts#L124 Ideally we should refactor this, likely by moving it into tinlake.js, but that's a larger item to tackle.

jpangelle:
### Description

This pull request implements the [`multicall.js`](https://github.com/makerdao/multicall.js) library for the `/portfolio` page in order to get real time data rather than just the daily updates from the subgraph.

I used `any[]` type on [here](https://github.com/centrifuge/apps/compare/use-multicall-on-portfolio#diff-0a9ef8b7e0d47f2972a9b22bf5c91b8a05e3e4cdec0f95c7f242c370ac075dbaR106) and [here](https://github.com/centrifuge/apps/compare/use-multicall-on-portfolio#diff-0a9ef8b7e0d47f2972a9b22bf5c91b8a05e3e4cdec0f95c7f242c370ac075dbaR120) because there seems to be a problem with the way that `multicall.js` ships their types. There is currently a [pull request](https://github.com/makerdao/multicall.js/pull/29) raised on the `multicall.js` repo to fix the issue.

[Staging Link](https://606f25b98533b600088856a2--tinlake-kovan-staging.netlify.app/)

Closes #113

### DIFF
--- a/tinlake-ui/components/Header/index.tsx
+++ b/tinlake-ui/components/Header/index.tsx
@@ -160,7 +160,7 @@ const Header: React.FC<Props> = (props: Props) => {
                       <TokenLogo src={`/static/DAI.svg`} />
                       <Box>
                         <Holdings>
-                          {addThousandsSeparators(toPrecision(baseToDisplay(portfolio.totalValue, 22), 0))}
+                          {addThousandsSeparators(toPrecision(baseToDisplay(portfolio.totalValue, 18), 0))}
                         </Holdings>
                         <Desc>Portfolio Value</Desc>
                       </Box>
--- a/tinlake-ui/components/Header/index.tsx
+++ b/tinlake-ui/components/Header/index.tsx
@@ -53,11 +53,11 @@ const Header: React.FC<Props> = (props: Props) => {
   const dispatch = useDispatch()
 
   React.useEffect(() => {
-    if (address) dispatch(loadPortfolio(address))
+    if (address) dispatch(loadPortfolio(address, props.ipfsPools))
   }, [])
 
   React.useEffect(() => {
-    if (address) dispatch(loadPortfolio(address))
+    if (address) dispatch(loadPortfolio(address, props.ipfsPools))
   }, [address])
 
   const connectAccount = async () => {
@@ -160,7 +160,7 @@ const Header: React.FC<Props> = (props: Props) => {
                       <TokenLogo src={`/static/DAI.svg`} />
                       <Box>
                         <Holdings>
-                          {addThousandsSeparators(toPrecision(baseToDisplay(portfolio.totalValue, 18), 0))}
+                          {addThousandsSeparators(toPrecision(baseToDisplay(portfolio.totalValue, 22), 0))}
                         </Holdings>
                         <Desc>Portfolio Value</Desc>
                       </Box>
--- a/tinlake-ui/ducks/portfolio.ts
+++ b/tinlake-ui/ducks/portfolio.ts
@@ -98,6 +98,11 @@ export function loadPortfolio(
 
     const watcher = createWatcher(watches, multicallConfig)
 
+    /*
+     * matches the token id's in tokenBalance with the token id's
+     * from the multicall to find the correct value (price) and balance
+     * any[] type instead of ICall[] type until https://github.com/makerdao/multicall.js/pull/29 is merged
+     */
     const findAmount = (updates: any[], balance: TokenBalance, infoType: 'price' | 'balance') => {
       const updateBalance = updates.find((update) => {
         const [tokenId, type] = update.type.split('-')
@@ -113,13 +118,17 @@ export function loadPortfolio(
 
     // any[] type instead of ICall[] type until https://github.com/makerdao/multicall.js/pull/29 is merged
     watcher.batch().subscribe((updates: any[]) => {
-      const newData = tokenBalances.map((balance: TokenBalance) => ({
+      /*
+       * overwrites the values in tokenBalances that were retrieved from
+       * the subgraph with the values from the multicall updates
+       */
+      const tokenBalancesUpdates = tokenBalances.map((balance: TokenBalance) => ({
         ...balance,
         value: findAmount(updates, balance, 'price'),
         balance: findAmount(updates, balance, 'balance'),
       }))
 
-      dispatch({ data: newData, type: RECEIVE_PORTFOLIO })
+      dispatch({ data: tokenBalancesUpdates, type: RECEIVE_PORTFOLIO })
     })
 
     watcher.start()
--- a/tinlake-ui/ducks/portfolio.ts
+++ b/tinlake-ui/ducks/portfolio.ts
@@ -122,13 +122,13 @@ export function loadPortfolio(
        * overwrites the values in tokenBalances that were retrieved from
        * the subgraph with the values from the multicall updates
        */
-      const tokenBalancesUpdates = tokenBalances.map((balance: TokenBalance) => ({
+      const updatedTokenBalances = tokenBalances.map((balance: TokenBalance) => ({
         ...balance,
         value: findAmount(updates, balance, 'price'),
         balance: findAmount(updates, balance, 'balance'),
       }))
 
-      dispatch({ data: tokenBalancesUpdates, type: RECEIVE_PORTFOLIO })
+      dispatch({ data: updatedTokenBalances, type: RECEIVE_PORTFOLIO })
     })
 
     watcher.start()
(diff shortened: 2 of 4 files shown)


tinlake-ui/ducks/portfolio.ts
```diff
@@ -1,9 +1,18 @@
...
 import { Action, AnyAction } from 'redux'
 import { ThunkAction } from 'redux-thunk'
+import config, { IpfsPools } from '../config'
 import Apollo from '../services/apollo'
 
+const multicallConfig = {
+  rpcUrl: config.rpcUrl,
+  multicallAddress: config.multicallContractAddress,
+  interval: 60000,
+}
```
jpangelle, line 14: Used the same `multicallConfig` used in [`ducks/pool.ts`](https://github.com/centrifuge/apps/blob/main/tinlake-ui/ducks/pool.ts#L12-L16)

tinlake-ui/components/Header/index.tsx
```diff
@@ -160,7 +160,7 @@ const Header: React.FC<Props> = (props: Props) => {
                       <TokenLogo src={`/static/DAI.svg`} />
                       <Box>
                         <Holdings>
-                          {addThousandsSeparators(toPrecision(baseToDisplay(portfolio.totalValue, 18), 0))}
+                          {addThousandsSeparators(toPrecision(baseToDisplay(portfolio.totalValue, 22), 0))}
```
jpangelle, line 163: Had to change various bases

tinlake-ui/components/Header/index.tsx
```diff
@@ -53,11 +53,11 @@ const Header: React.FC<Props> = (props: Props) => {
   const dispatch = useDispatch()
 
   React.useEffect(() => {
-    if (address) dispatch(loadPortfolio(address))
+    if (address) dispatch(loadPortfolio(address, props.ipfsPools))
```
jpangelle, line 56: `Portfolio Value` in the header will reflect real time data.

image

tinlake-ui/components/Header/index.tsx
```diff
@@ -53,11 +53,11 @@ const Header: React.FC<Props> = (props: Props) => {
...
   }, [])
 
   React.useEffect(() => {
-    if (address) dispatch(loadPortfolio(address))
+    if (address) dispatch(loadPortfolio(address, props.ipfsPools))
   }, [address])
```
jpangelle, line 61: `Portfolio Value` in the header will reflect real time data.

image

tinlake-ui/ducks/portfolio.ts
offerijns: Since we're using the same constant in multiple files now, what do you think about moving this entire block to `config.ts` and exporting it from there?

tinlake-ui/ducks/portfolio.ts
```diff
@@ -55,9 +64,73 @@ export default function reducer(
...
+
+    // any[] type instead of ICall[] type until https://github.com/makerdao/multicall.js/pull/29 is merged
+    watcher.batch().subscribe((updates: any[]) => {
+      /*
+       * overwrites the values in tokenBalances that were retrieved from
+       * the subgraph with the values from the multicall updates
+       */
+      const updatedTokenBalances = tokenBalances.map((balance: TokenBalance) => ({
+        ...balance,
+        value: findAmount(updates, balance, 'price'),
```
offerijns, line 127: I think you're confusing the value and price terms here. Generally, `value = price * balance`. So I think this one should be called `price`, right?

tinlake-ui/components/Header/index.tsx
offerijns: This could become a bit confusing if we are going to load the data in other places at some point. Generally, bases are either 18 or 27, depending on what they are in the contracts. If you have to multiply values and they change, I think it's best to immediately divide back to the right base. So in this case, I think `portfolio.totalValue` should be divided by `10**4` to get back to base 18 in the portfolio duck already, so here you can just display it as base 18. Does that make sense?

offerijns (approved):
Left a few small comments, but overall it looks great!

tinlake-ui/ducks/portfolio.ts
```diff
@@ -55,9 +64,73 @@ export default function reducer(
...
+        returns: [[`${pool.addresses.JUNIOR_TOKEN}-balance`, toBN]],
+      },
+      {
+        target: pool.addresses.SENIOR_TOKEN,
+        call: ['balanceOf(address)(uint256)', address],
+        returns: [[`${pool.addresses.SENIOR_TOKEN}-balance`, toBN]],
+      },
+    ])
+
+    const watcher = createWatcher(watches, multicallConfig)
```
lucasvo, line 99: Ideally we shouldn't create a watcher here but reuse one global watcher. This is creating another request for no reason.

lucasvo: see discussion on slack, perhaps the overhead of managing one per page is not worth it.

jpangelle: Read the Slack convo. Looks like it was decided that making the watchers global would be a pre-mature optimization. Something we can revisit in the future. Is that right?

tinlake-ui/ducks/portfolio.ts
jpangelle: I did consider that but since it is only occurring twice and isn't really logic heavy, I sort feel like it is ok to just have this as duplicate code for now.

tinlake-ui/ducks/portfolio.ts
jpangelle: Ok, I looked into this and I think there's some confusion on the use of the term `value`. The [`TokenBalance` type](https://github.com/centrifuge/apps/blob/main/tinlake-ui/ducks/portfolio.ts#L11-L20) has a `value` property that I think is representing the `price`. You can see the GraphQL [here](https://github.com/centrifuge/apps/blob/4547082ce44d99303af748aa503fec1f58501ee8/tinlake-ui/services/apollo/index.ts#L462-L473). But in the table on `/portfolio`, the terms are used differently. `Current Price` describes the token price. `Current Value` describes the `balance * price`
image

Basically, the code is calling the property "value" while the UI is calling it "price". How do you think we should reconcile this?

tinlake-ui/components/Header/index.tsx
jpangelle: That makes sense.

tinlake-ui/ducks/portfolio.ts
jpangelle: Hold on that thought. I think I actually confused myself.

jpangelle: Ok yeah, I cleared it up. Disregard what I said above.

tinlake-ui/ducks/portfolio.ts
```diff
@@ -55,9 +64,81 @@ export default function reducer(
...
+  address: string,
+  ipfsPools: IpfsPools
+): ThunkAction<Promise<void>, PortfolioState, undefined, Action> {
   return async (dispatch) => {
     const tokenBalances = await Apollo.getPortfolio(address)
-    dispatch({ data: tokenBalances, type: RECEIVE_PORTFOLIO })
+
+    const toBN = (val: BigNumber) => new BN(val.toString())
+
+    const watches = ipfsPools.active.flatMap((pool) => [
```
offerijns, line 76:
```suggestion
    const watchers = ipfsPools.active.flatMap((pool) => [
```

tinlake-ui/ducks/portfolio.ts
```diff
@@ -55,9 +64,81 @@ export default function reducer(
...
+        returns: [[`${pool.addresses.JUNIOR_TOKEN}-balance`, toBN]],
+      },
+      {
+        target: pool.addresses.SENIOR_TOKEN,
+        call: ['balanceOf(address)(uint256)', address],
+        returns: [[`${pool.addresses.SENIOR_TOKEN}-balance`, toBN]],
+      },
+    ])
+
+    const watcher = createWatcher(watches, multicallConfig)
```
offerijns, line 99:
```suggestion
    const watcher = createWatcher(watchers, multicallConfig)
```

tinlake-ui/ducks/portfolio.ts
offerijns: Yes that's right!

offerijns (approved):
One small typo, otherwise it looks great!

PR #180, opened 2021-04-08
docs: tinlake-ui release rfc

jpangelle:
### Description

This pull request contains an RFC (request for comments) regarding a proposal for a slight change to the `tinlake-ui` release process.

PR #191, opened 2021-04-22
ci: implement netlify github action

jpangelle:
### Description

This pull request accomplishes the first step in the [tinlake-ui release rfc](https://github.com/centrifuge/apps/pull/180). Some prerequisites for this to work:

1. Create new domains on Netlify:
- `staging.tinlake.centrifuge.io` ✅
- `dev.tinlake.centrifuge.io` ✅

2. Rename Github secret `TINLAKE_UI_NETLIFY_SITE_ID` to `TINLAKE_UI_PROD_NETLIFY_SITE_ID` ✅

3. Add site id's to Github secrets:
- `TINLAKE_UI_DEV_NETLIFY_SITE_ID`: `<dev.tinlake.centrifuge.io site id>` ✅
- `TINLAKE_UI_KOVAN_STAGING_NETLIFY_SITE_ID`: `<kovan.staging.tinlake.centrifuge.io site id>` ✅
- `TINLAKE_UI_MAINNET_STAGING_NETLIFY_SITE_ID`: `<staging.tinlake.centrifuge.io site id>` ✅
- `TINLAKE_UI_PROD_NETLIFY_SITE_ID`: `<tinlake.centrifuge.io site id>` ✅


.github/workflows/tinlake-ui-deploy-to-staging.yml
```diff
@@ -0,0 +1,123 @@
+name: 'tinlake-ui Deploy to Staging'
```
jpangelle, line 1: This workflow has two different deploy steps:

1. Deploy to `kovan.staging.tinlake.centrifuge.io` pointing at Kovan testnet and upload to IPFS
2. Deploy to `staging.tinlake.centrifuge.io` pointing at mainnet

.github/workflows/tinlake-ui-pull-request.yml
```diff
@@ -0,0 +1,59 @@
...
+          NEXT_PUBLIC_PORTIS_KEY: 'bc9e2922-2ebd-4e2b-86f6-7c7855bdf07f'
+          NEXT_PUBLIC_REWARDS_TREE_URL: 'https://storage.googleapis.com/rad-rewards-trees-kovan-staging/latest.json'
+          NEXT_PUBLIC_RPC_URL: 'https://kovan.infura.io/v3/f9ba987e8cb34418bb53cdbd4d8321b5'
+          NEXT_PUBLIC_TINLAKE_DATA_BACKEND_URL: 'https://api.thegraph.com/subgraphs/name/jennypollack/tinlake-rewards-staging'
+          NEXT_PUBLIC_TRANSACTION_TIMEOUT: '3600'
+
+      - name: Deploy To Netlify
+        uses: nwtgck/actions-netlify@v1.1
+        with:
+          alias: pr-${{ github.event.pull_request.number }}
```
jpangelle, line 50: Uses an alias for the preview link so that subsequent commits to a pull request will deploy to the same url. That way, if you share the preview link to someone and then you push changes, it will deploy to the same url and you will not have to send them an updated link. Example `dev.tinlake.centrifuge.io/pr-183`

offerijns: Wow, this will be really useful!

offerijns (approved):
This looks great! One thing I did note: this means we'll also be managing the env vars for all the environments in the github workflows, rather than in the Netlify UI like we do now. I wonder if we can come up with a better approach for that. E.g. I was just looking at this: https://github.com/marketplace/actions/dotenv-action. Maybe we could create separate `.env.mainnet` and `.env.kovan` files, and then use this action to pull these into the workflows? Although it's also not ideal because we'd still have to overwrite some of them as they are secrets. Maybe we should look into this a bit more and figure out the best option.

.github/workflows/tinlake-ui-deploy-to-dev.yml
```diff
@@ -0,0 +1,61 @@
...
+          NEXT_PUBLIC_MULTICALL_CONTRACT_ADDRESS: '0x2cc8688c5f75e365aaeeb4ea8d6a480405a48d2a'
+          NEXT_PUBLIC_ONBOARD_API_HOST: 'https://onboard-api.amber.centrifuge.io/'
+          NEXT_PUBLIC_POOL_REGISTRY: '0x8FE85CeAe6157C1dfcDD1c5ec99361c9722d97de'
+          NEXT_PUBLIC_POOLS_CONFIG: 'kovanStaging'
+          NEXT_PUBLIC_POOLS_IPFS_HASH_OVERRIDE: 'QmYAgaJi8YGok3Q26RLBqEq7fhPtcLRBCAndVXJ6KWnQCg'
+          NEXT_PUBLIC_PORTIS_KEY: 'bc9e2922-2ebd-4e2b-86f6-7c7855bdf07f'
+          NEXT_PUBLIC_REWARDS_TREE_URL: 'https://storage.googleapis.com/rad-rewards-trees-kovan-staging/latest.json'
+          NEXT_PUBLIC_RPC_URL: 'https://kovan.infura.io/v3/f9ba987e8cb34418bb53cdbd4d8321b5'
+          NEXT_PUBLIC_TINLAKE_DATA_BACKEND_URL: 'https://api.thegraph.com/subgraphs/name/jennypollack/tinlake-rewards-staging'
+          NEXT_PUBLIC_TRANSACTION_TIMEOUT: '3600'
```
philipstanislaus, line 47: Is these a better way to store those? E. g. using [environments/secrets of GitHub](https://docs.github.com/en/actions/reference/environments)? The main issue doing it like this here is that we need a separate workflow for each deployment environment, which might be fine actually... Not sure, what do you think?

philipstanislaus (approved):
Awesome, love this new process!

philipstanislaus:
> This looks great! One thing I did note: this means we'll also be managing the env vars for all the environments in the github workflows, rather than in the Netlify UI like we do now. I wonder if we can come up with a better approach for that. E.g. I was just looking at this: https://github.com/marketplace/actions/dotenv-action. Maybe we could create separate `.env.mainnet` and `.env.kovan` files, and then use this action to pull these into the workflows? Although it's also not ideal because we'd still have to overwrite some of them as they are secrets. Maybe we should look into this a bit more and figure out the best option.

Just realized I wrote almost the same in my review 🙈

jpangelle:
@offerijns @philipstanislaus I researched ways to abstract the environment variables to a centralized location for easier maintenance and there does not seem to be a trivial way to do this. One approach is to leverage [Github Environments](https://docs.github.com/en/actions/reference/environments#referencing-an-environment), but this feature is not accessible by contributors who are not admins on the repo. Ideally, the environment variables would be contained in a file like Jeroen laid out (`.env.kovan` & `.env.mainnet`). Since we don't foresee these environment variables changing very often, I think it is best we just move forward with the workflow config as is it now. Perhaps we can optimize in the future if this config proves to be painful.

philipstanislaus:
Sounds good to me, thanks @jpangelle !

PR #198, opened 2021-04-29
fix: hook use in thunk

jpangelle:
Was getting this runtime error since `useQueryDebugEthAddress` was being used inside a `thunk` which goes against React's [rules of hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-from-react-functions) can only be used in function components.

I refactored to move the implementation to the `Auth` component which passes the hook's return data to the `thunk`.

Also slipped in a few ci tweaks 😉

### DIFF
--- a/.github/workflows/tinlake-ui-deploy-to-dev.yml
+++ b/.github/workflows/tinlake-ui-deploy-to-dev.yml
@@ -38,7 +38,7 @@ jobs:
           key: ${{ runner.os }}-nextjs-${{ hashFiles('**/package-lock.json') }}
 
       - name: Setup Node
-        uses: actions/setup-node@v2-beta
+        uses: actions/setup-node@v2
         with:
           node-version: '12'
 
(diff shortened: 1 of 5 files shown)


.github/workflows/tinlake-ui-deploy-to-dev.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           key: ${{ runner.os }}-nextjs-${{ hashFiles('**/package-lock.json') }}
```
jpangelle, line 38: ignore the ci files. I slipped in these tweaks to this PR. 😄

tinlake-ui/components/Auth/index.tsx
```diff
@@ -3,35 +3,28 @@ import * as React from 'react'
 import { connect } from 'react-redux'
```
jpangelle, line 3: Refactored from a class component to a functional component so to make easier to use the `useQueryDebugEthAddress` hook and then pass the result to the `load` action.

offerijns (approved):
👍

philipstanislaus (commented):
Awesome, thanks @jpangelle ! Not sure how this slipped my tests... 

🎉

PR #218, opened 2021-05-06
fix: re-prettier code

jpangelle:
### Description

Our code formatting recently got out of sync which was causing the `check-format` ci job to fail. I ran `yarn format:run` to reformat the codebase. All of the code changes in the pull request are auto-formatting fixes.


offerijns (approved):
I think that's my bad, thanks for fixing it 👍

PR #225, opened 2021-05-20
feat: add AO rewards to investor rewards

jpangelle:
### Description

This pull requests adds the AO rewards to the investor rewards amount on the dashboard page.

**NOTE:** This functionality will not work on `Kovan` or `mainnet` until the latest version of the subgraph that contains the [toDateAORewardAggregateValue](https://github.com/centrifuge/tinlake-subgraph/blob/39558c47e892ac7b2825c6a051a9925567d59a04/schema.graphql#L187) property is deployed. Otherwise, the dashboard will render with `0.0000 CFG`:
image

### DIFF
--- a/tinlake-ui/components/PoolsMetrics/index.tsx
+++ b/tinlake-ui/components/PoolsMetrics/index.tsx
@@ -28,7 +28,16 @@ const PoolsMetrics: React.FC<Props> = (props: Props) => {
     dispatch(maybeLoadRewards())
   }, [])
 
-  const totalRewardsEarned = baseToDisplay(rewards.data?.toDateRewardAggregateValue || '0', 18)
+  const AORewardsEarned = rewards.data?.toDateAORewardAggregateValue
+  const investorRewardsEarned = rewards.data?.toDateRewardAggregateValue
+
+  const totalRewardsEarned = React.useMemo((): string => {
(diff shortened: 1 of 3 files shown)


tinlake-ui/components/PoolsMetrics/index.tsx
```diff
@@ -28,7 +28,16 @@ const PoolsMetrics: React.FC<Props> = (props: Props) => {
...
+  const AORewardsEarned = rewards.data?.toDateAORewardAggregateValue
+  const investorRewardsEarned = rewards.data?.toDateRewardAggregateValue
+
+  const totalRewardsEarned = React.useMemo((): string => {
+    if (AORewardsEarned && investorRewardsEarned) {
+      return baseToDisplay(investorRewardsEarned.add(AORewardsEarned), 18)
+    }
+
+    return '0'
+  }, [AORewardsEarned, investorRewardsEarned])
```
jpangelle, line 40: This logic assumes that `rewards.data.toDateAORewardAggregateValue` and `rewards.data.toDateRewardAggregateValue` will always return some truthy value otherwise one of them is falsy (`undefined`, `0`), the function will return `'0'`. I am assuming at worst, the value would be `'0'` which would be truthy so it would resolve correctly.

offerijns (approved):
This is great, thanks!

PR #242, opened 2021-06-11
chore: update readme and ci

jpangelle:
### Description

This pull request:
- updates the `README.md` to remove the Netlify badges since they are not compatible with using a homebaked CI setup
- adds own workflow path to `deploy-to-dev` and `pull-request` workflows so that deploys are triggered when only `env`s are updated in a code change


offerijns (approved):
👍

Issue #235, opened 2021-06-09
Remove/Update `Close epoch` implementation


denniswell:
MUST HAVE
✅ (1) Remove button/functionality for public
✅ (2) Make available to stakeholders:
- The button `Close epoch` should only be displayed to:
    - All Pool admins (on new pools with the new pool admin contract)
    - Behind a feature flag `/investments&show_close_epoch=true`

NICE TO HAVE:
(3) Update flow for stakeholders
Background, there are some issues with the current implementation. 
(i) the UI doesn't update after a transaction --> Please make sure it does 
(ii) a few UI UX improvements can be made in the flow
Background: Current transaction flow
(1) When epoch minimum duration is passed --> Close epoch
    - if it executes automatically, flow ends here
    - if it does not automatically execute:
(2) `Submit solution` (run's the solver and submits a solution
(3) `Execute epoch` 

UPDATES on current flow
(a) While epoch is ongoing, please show `Close epoch` button but deactivate it (currently it is hidden)

PR #249, opened 2021-06-15
feat: conditionally render "Close epoch" button

jpangelle:
### Description/

This pull request hides the `Close epoch` button on the investment page unless the user is either the pool admin or uses query params (`?show_close_epoch=true`) in the url.

[Preview link with query params](https://pr-249--dev-tinlake.netlify.app/pool/0x25dF507570c8285E9c8E7FFabC87db7836850dCd/new-silver-2/investments?show_close_epoch=true)

Addresses the `MUST HAVE`s of #235

### DIFF
--- a/tinlake-ui/containers/Investment/View/EpochOverview.tsx
+++ b/tinlake-ui/containers/Investment/View/EpochOverview.tsx
@@ -23,7 +23,7 @@ interface Props extends TransactionProps {
 }
 
 const EpochOverview: React.FC<Props> = (props: Props) => {
-  const { query } = useRouter()
+  const router = useRouter()
 
   const pool = useSelector<any, PoolState>((state) => state.pool)
   const poolData = pool?.data as PoolData | undefined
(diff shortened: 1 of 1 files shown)


tinlake-ui/containers/Investment/View/EpochOverview.tsx
```diff
@@ -44,6 +47,34 @@ const EpochOverview: React.FC<Props> = (props: Props) => {
 
   const [open, setOpen] = React.useState(false)
 
+  const EpochButton = React.useCallback(() => {
+    switch (epochData?.state) {
+      case 'can-be-closed':
+        if (poolData?.isPoolAdmin || query.show_close_epoch === 'true') {
```
offerijns, line 53: This wasn't clear in the specs, but this should apply also to the `Submit a solution` and `Execute epoch` buttons.

tinlake-ui/containers/Investment/View/EpochOverview.tsx
```diff
@@ -44,6 +47,34 @@ const EpochOverview: React.FC<Props> = (props: Props) => {
 
   const [open, setOpen] = React.useState(false)
 
+  const EpochButton = React.useCallback(() => {
```
offerijns, line 50: Nice way to separate this 👍

tinlake-ui/containers/Investment/View/EpochOverview.tsx
```diff
@@ -22,6 +23,8 @@ interface Props extends TransactionProps {
 }
 
 const EpochOverview: React.FC<Props> = (props: Props) => {
+  const { query } = useRouter()
```
offerijns, line 26: I personally tend to prefer not destructuring too much, as the context can get lost. E.g. if you see `query` somewhere else in a large component, you don't really know what it's referring to, which is different from if it wasn't destructured and it was referred to by `router.query`. I'm curious what you think about this.

offerijns (changes requested):

tinlake-ui/containers/Investment/View/EpochOverview.tsx
jpangelle: Ah makes sense. Will add.

tinlake-ui/containers/Investment/View/EpochOverview.tsx
jpangelle: Cool, whatever makes things more readable, I am on board. I will refactor to be more verbose.

offerijns (approved):
Perfect!

PR #283, opened 2021-07-21
Bugfix/subgraph error

### LINKED ISSUE #229: Fix tinlake-ui not loading when the subgraph is down
When the subgraph can not be loaded, the entire dApp crashes. Instead it should fail more gracefully. Most of the subgraph data for `tinlake-ui` is not based on the subgraph so the app should still be able to load up for the most part.

Current error: 
image

onnovisser:
Fixes app crash when subgraph is down.
Also adds a simple loading animation to the placeholders ([loom](https://www.loom.com/share/29c99a38255847b78581a3992638407c))

Closes #229


jpangelle (approved):
Works for me! Nice! 🎸

PR #284, opened 2021-07-22
always show pending transations

### LINKED ISSUE #143: Display Pending transaction toasts until tx confirmed
Background
Pending tx toasts currently disappear after five seconds and are only visible after opening the wallet component.

Change:
Display toasts until tx confirmed plus 3 seconds

onnovisser:
Don't hide pending transactions after a timeout, but keep them visible until they succeed or fail.
Closes #143


jpangelle (approved):
💯 LGTM!

Issue #144, opened 2021-02-21
Improve locking NFT flow


denniswell:
Ready for dev:
- [x] (i) Display of Lock NFT button
This is confusing for users other than the AO. Only the borrower address should see it, so
Only display lock NFT button when (i) accessed with URL `lockNFT=true` or connected address as a proxy for this pool
- [x] Confirmation for locked NFT
AOs usually stay on the `Lock NFT` page and are confused when nothing happens. When they do stay on the page and an NFT is succesfully locked, we should at least show a message box: "Your NFT is succesfully locked. Please proceed to the [Asset list](link to Asset tab) to finance this NFT. Doesnt need to be pretty, only AOs see it, similar to kovan boxes is fine
- [x] Hide Finance/repay buttons instead of deactivate
I know we had the discussion and I was wrong. After seeing this in action with a few handholdings, it is much clearer if
Until the first financing - Only the Financing button is show, Repay is hidden
After the first financing - Financing button is hidden, value in input field is 0 and only repay button is shown
(iv) Rename and move`close` loan button
Rename `Close` button to `Unlock NFT`. Can remain secondary button. Only display before first borrowing, otherwise hide. The button should be moved into top right corner, where the 'Lock NFT' button is also positioned

lucasvo:
Instead of requiring a feature flag in the URL for "lockNFT" why don't we check if the connected address has a tinlake borrower proxy and can use the pool?

denniswell:
> Instead of requiring a feature flag in the URL for "lockNFT" why don't we check if the connected address has a tinlake borrower proxy and can use the pool?

This is a hen and egg problem for the first NFT, as to create the borrower proxy you need the "lock NFT" button. Thus, for first financing --> feature flag, afterwards --> automatically displayed if proxy exists...

PR #301, opened 2021-07-30
Asset tab UI tweaks

### LINKED ISSUE #109: Add: Success/Confirmation NFT locked
When AOs Lock NFT they stay on the page for 95% of the time, however they get no confirmation whether that was succesful
In case they stay on the page until tx confirmation, we should show a confirmation message (our Kovan desing is fine for now, only AOs see it anyways) with
"Your NFT has been successfully locked. Please proceed to finance the asset [here](link to Asset details page)
If they don't stay on the page, all good, then the toast is enough

onnovisser:
Fixes (i), (ii) and (iii) of #144 

(i) Show "Lock NFT" button only when account is pool admin, has a proxy, or `lockNFT` query param is set.
(ii) Adds success feedback for locking NFT
(iii) Hide "Repay" button when you can borrow, hide "Finance" button when you can't.

Closes #109

### DIFF
--- a/tinlake-ui/pages/pool/[root]/[slug]/assets/index.tsx
+++ b/tinlake-ui/pages/pool/[root]/[slug]/assets/index.tsx
@@ -2,7 +2,9 @@ import { Box, Button, Heading } from 'grommet'
 import { GetStaticProps } from 'next'
 import { WithRouterProps } from 'next/dist/client/with-router'
 import Head from 'next/head'
+import { useRouter } from 'next/router'
 import * as React from 'react'
+import { useSelector } from 'react-redux'
 import Auth from '../../../../../components/Auth'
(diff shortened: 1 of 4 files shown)


tinlake-ui/pages/pool/[root]/[slug]/assets/index.tsx
```diff
@@ -21,60 +25,65 @@ interface Props extends WithRouterProps {
   ipfsPools: IpfsPools
 }
 
-class LoanListPage extends React.Component<Props> {
-  render() {
-    const { pool, ipfsPools } = this.props
+const LoanListPage: React.FC<Props> = (props) => {
```
Comment (offerijns) [inline tinlake-ui/pages/pool/[root]/[slug]/assets/index.tsx:28]: 👍

offerijns (approved):

PR #304, opened 2021-08-02
Mobile improvements

onnovisser:
Makes some small changes to make things look a bit better on mobile. Mostly making cards stack vertically on small screens.

image
image
image


offerijns (approved):
LGTM!

PR #312, opened 2021-08-03
add underlines to tooltips

### LINKED ISSUE #289: Add visual clue to content elements bearing a tooltip
#### Background
Labels and value with a tooltip providing further information currently do not offer any affordance of the useful and valuable information. This makes the information not easy to discover. Tooltips are important for users learn and understand Tinlake on their own.

Example:
Asset maturity, DROP APR and Pool value contain tooltips – Asset type does not.

#### Approach to solution
Text elements carrying a tooltip should provide a unobtrusive affordance so users know where to expect a tooltip. The tooltip format must not compete with the actual content, and must not be confused with a link. The dotted underline(similar to the [HTML abbreviation element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr)) is an established solution to this problem.

The tooltip format should be consistently applied **to all text elements in the Tinlake UI**.

Solution example:

[Figma](https://www.figma.com/file/EWziKt1qCFdHGhWdnkrcKf/Tinlake-UI?node-id=5644%3A16998)

onnovisser:
Adds underlines to most tooltips.

image

image

Closes #289


tinlake-ui/components/Tooltip/index.tsx
```diff
@@ -1,45 +1,65 @@
...
-    <>{props.children}</>
+  let title
+  let link
+  let description
+
+  if ('id' in props) {
+    ;({ title, link, description } = tooltips[props.id] as Tooltip)
+  } else {
+    ;({ title, link, description } = props)
+  }
```
jpangelle, line 45: 🙈

jpangelle (approved):
nice!

tinlake-ui/components/Tooltip/index.tsx
onnovisser: updated 😏

PR #314, opened 2021-08-04
Add investment capacity to pool detail

### LINKED ISSUE #306: Display investment capacity in pool detail
### Background
The pool detail currently shows the following status:
- `Oversubscribed`
- `Upcoming`

*Fig. Oversubscribed pool status in detail view*

*Fig. Upcoming pool status in detail view*

The investment capacity of a pool has been added to the pool list on the [dashboard](https://centrifuge.hackmd.io/OMfFMGTMQn6uMYfG9Af4nA?both). This information is currently missing in the pool detail views.

### Approach
The investment capacity is added as status feedback to the pool detail, so users can see the pool's capacity in context, without relying on memory.

#### Rules
- The status uses the same format like on the dashboard
- The amount is rounded to 1,000 DAI, like on the dashboard
- The status uses the same position as the other status (Oversubscribed, Upcoming) – only one status applies at a time

[Figma](https://www.figma.com/file/EWziKt1qCFdHGhWdnkrcKf/Tinlake-UI?node-id=5776%3A20304)

onnovisser:
Add investment capacity to pool detail page.

image

Closes #306


tinlake-ui/components/PoolCapacityLabel/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+import BN from 'bn.js'
+import * as React from 'react'
+import styled from 'styled-components'
+import { PoolData } from '../../ducks/pools'
+
+interface Props {
+  pool: PoolData
+}
+
+export const PoolCapacityLabel: React.FC<Props> = ({ pool: p }) => {
```
jpangelle, line 11: I think it might be better to use proper `if` statements here instead of a series of nested ternaries so that we can preserve readability:

```ts
export const PoolCapacityLabel: React.FC<Props> = ({ pool: p }) => {
  const {
    assetValue,
    capacity,
    currency,
    isArchived,
    isOversubscribed,
    reserve,
  } = p;

  const isUpcoming =
    p.isUpcoming ||
    (!assetValue && !reserve) ||
    (assetValue?.isZero() && reserve?.isZero());

  if (isUpcoming) {
    return <Label blue>Upcoming</Label>;
  }

  if (isArchived) {
    return <Label>Archived</Label>;
  }

  if (isOversubscribed) {
    return <Label orange>Oversubscribed</Label>;
  }

  return (
    <Label green>
      {addThousandsSeparators(
        toPrecision(baseToDisplay(capacity || new BN(0), 21), 0),
      )}
      K {currency}
    </Label>
  );
};

```

jpangelle (approved):
👍 
one suggestion but not a blocker

tinlake-ui/components/PoolCapacityLabel/index.tsx
onnovisser: Yeah, much better. Updated!

PR #321, opened 2021-08-06
Fix portfolio balance

onnovisser:
Portfolio balance wasn't counting pending token redemptions. 
Also now using multicall instead of the subgraph for portfolio balances, so it doesn't lag behind.


offerijns (approved):
Looks good 👍  I'm not sure we really need lodash, but we've not been particularly selective about existing dependencies either, so it would be fine to add it if you feel it's helpful!

onnovisser:
Yeah, was a bit on the fence re lodash. I do like to keep things slim and we could still replace it if we don't end up using more of it.

PR #325, opened 2021-08-10
Order pools on Dashboard by Investment Capacity

### LINKED ISSUE #322: Order pools on Dashboard by Investment Capacity
ACs:
Now that we have Investment Capacity shown (correctly and reliably it seems for now) on the dashboard, we should order pools by it, so the ones with the greatest capacity are on top (and not only the largest) if possible.
Ordering would be
- InvCap (largest to smallest)
- Oversubscribed (secondary ordering then pool size if possible/required)
- Upcoming

TC:
TBD

onnovisser:
Now hides the pool list until the data from the subgraph and multicall have been fetched, because it needs the additional data to sort and otherwise the pools in the list jump around when it eventually gets sorted

Closes #322


tinlake-ui/ducks/pools.ts
```diff
@@ -216,124 +206,128 @@ export function loadPools(pools: IpfsPools): ThunkAction<Promise<void>, { pools:
...
+        },
+      ] = await Promise.all([Apollo.getPools(pools), aggregate(calls, multicallConfig)])
+
+      const updatesPerPool: any = {}
+      Object.entries(multicallData).forEach(([type, value]) => {
+        const poolId = type.split('.')[0]
+        const key = type.split('.')[1]
+        if (!(poolId in updatesPerPool)) updatesPerPool[poolId] = {}
+        updatesPerPool[poolId][key] = value
+      })
```
jpangelle, line 224: I was toying around with reduce so that this can be a "pure" operation rather than declaring an object with `any` and then adding stuff to it. Do you think this any better?

```js
const updatesPerPool = Object.entries(multicallData).reduce(
  (updates, [type, value]) => {
    const [poolId, key] = type.split('.');

    if (!(poolId in updates)) {
      updates[poolId] = {};
    }

    updates[poolId][key] = value;

    return updates;
  },
  {},
);
```

jpangelle (approved):
one comment, no blockers 👍

tinlake-ui/ducks/pools.ts
onnovisser: Nice yeah would be better. Updated! Still needed `any` inside the loop though but at least the result is typed better now.
```
const updatesPerPool: { [key: string]: State } = Object.entries(multicallData).reduce(
        (updates: any, [type, value]) => {
          const [poolId, key] = type.split('.')

          if (!(poolId in updates)) {
            updates[poolId] = {}
          }

          updates[poolId][key] = value

          return updates
        },
        {}
      )
```

PR #326, opened 2021-08-11
Prevent borrower from Locking NFT that is not in his wallet

### LINKED ISSUE #67: Prevent borrower from Locking NFT that is not in his wallet
This is important and should be fixed soon.

We currently allow the borrower to try to lock any existing NFT regardless of it being in his wallet or not. This leads to failed transactions (without appropriate error message) for the borrower. This happens regularly, when the subgraph is not quick enough to update the asset list, the borrower can't see the financing after locking the NFT, tries again and it fails because it was already removed from the wallet.

Currently the 'Lock NFT' button is activated once NFT registry and NFT ID are entered in the correct format. We should also add a check whether the NFT is actually held by the account that wants to finance it. If the NFT ID has the correct expected format but is not held by the account/address then we should show an error message (we have the input component that shows errors subline) `NFT not held in your wallet`.

onnovisser:
Disabled the Lock NFT button when the owner of the NFT isn't the connected address.
Also fixes the error messages that were not being displayed.

Closes #67

### DIFF
--- a/tinlake-ui/containers/Loan/Issue/index.tsx
+++ b/tinlake-ui/containers/Loan/Issue/index.tsx
@@ -7,7 +7,7 @@ import Alert from '../../../components/Alert'
 import NftData from '../../../components/NftData'
 import { PoolLink } from '../../../components/PoolLink'
 import { Pool } from '../../../config'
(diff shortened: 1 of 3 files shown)


offerijns (approved):
Nice!

PR #331, opened 2021-08-13
Automatically close pools if DROP redemptions can be taken out of reserve

### LINKED ISSUE #267: SOLVER BOT: Automatically close pools if DROP redemptions can be taken out of reserve
Currently pool epochs are not automatically closed by the bot if not all orders can be fully executed. This sometimes prevents DROP redemption orders that could be served from the reserve not being executed because, e.g. there are also DROP investments locked that cannot be executed because the pool is oversubscribed

We can easily add an additional case where the bot closes the epoch to the bot:
If DROP redemptions locked > 0 then
check that
(i) no TIN investments/redemptions are locked (we don’t wont do do any crazy stuff)
(ii) Locked redemptions < Pool reserve
Then close epoch

onnovisser:
Closes #267

### DIFF
--- a/tinlake-bot/src/tasks/closePools.ts
+++ b/tinlake-bot/src/tasks/closePools.ts
@@ -13,7 +13,7 @@ export const closePools = async (pools: PoolMap, provider: ethers.providers.Prov
   for (let pool of Object.values(pools)) {
     try {
       if (!pool.addresses) continue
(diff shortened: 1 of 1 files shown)


jpangelle (approved):
👍

offerijns (approved):

PR #342, opened 2021-08-19
Gateway: Redirect to login when session is expired

### LINKED ISSUE #206: Check Gateway timeout on inactivity
I think we may have changed the automatic Tinlake timeout on inactivity, I have remained logged in quite a while when testing last week. We should set it to 10min

onnovisser:
image

Redirects the user to the login page when the API returns a 401.
Refactors the httpclient to be passed via context, so it can access the history and auth context, in order to redirect and clear the auth data.

Also adds an client-side idle timeout that logs the user out after being idle for 10 minutes.

Closes #206


offerijns:
Closing since it's stale.

PR #357, opened 2021-09-02
feat: add goerli config support

offerijns:


onnovisser (commented):
Could you add a `.env.goerli-example` ?

PR #362, opened 2021-09-06
Fix management page breaking when pool address in URL is lowercase

### LINKED ISSUE #361: Pool management page doesn't load when pool address is lowercase
Some browsers don't handle case-sensitive URLs well, and turn those with uppercase characters in lowercase equivalents. This breaks the management page (but somehow not other pages?). We should convert to lowercase when comparing pool addresses from the url to the root contract in the pool configs.

onnovisser:
Closes #361

### DIFF
--- a/tinlake-ui/pages/pool/[root]/[slug]/management/index.tsx
+++ b/tinlake-ui/pages/pool/[root]/[slug]/management/index.tsx
@@ -60,25 +60,20 @@ const ManagementPage: React.FC<Props> = ({ pool, ipfsPools }) => {
 export async function getStaticPaths() {
   // We'll pre-render only these paths at build time.
   const pools = await loadPoolsFromIPFS()
-  const paths = pools.active.flatMap((pool) => [
-    {
-      params: { root: pool.addresses.ROOT_CONTRACT, slug: pool.metadata.slug },
-    },
-    {
-      params: { root: pool.addresses.ROOT_CONTRACT.toLowerCase(), slug: pool.metadata.slug },
-    },
-  ])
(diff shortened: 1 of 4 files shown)


tinlake-ui/pages/pool/[root]/[slug]/management/index.tsx
```diff
@@ -60,20 +60,25 @@ const ManagementPage: React.FC<Props> = ({ pool, ipfsPools }) => {
...
   const pools = await loadPoolsFromIPFS()
-  const paths = pools.active.map((pool) => ({
-    params: { root: pool.addresses.ROOT_CONTRACT, slug: pool.metadata.slug },
-  }))
+  const paths = pools.active.flatMap((pool) => [
+    {
+      params: { root: pool.addresses.ROOT_CONTRACT, slug: pool.metadata.slug },
+    },
+    {
+      params: { root: pool.addresses.ROOT_CONTRACT.toLowerCase(), slug: pool.metadata.slug },
```
Comment (alessioscalici) [inline tinlake-ui/pages/pool/[root]/[slug]/management/index.tsx:68]: Is this guaranteed to exist, and be of type string?

Comment (onnovisser) [reply tinlake-ui/pages/pool/[root]/[slug]/management/index.tsx:68]: yes, there are some contracts that are optional, but ROOT_CONTRACT is always there

alessioscalici (approved):

tinlake-ui/pages/pool/[root]/[slug]/management/index.tsx
Comment (alessioscalici) [reply tinlake-ui/pages/pool/[root]/[slug]/management/index.tsx:68]: thanks!

PR #372, opened 2021-09-13
Add ValuePairList and Value component

onnovisser:
Adds ValuePairList component with three different variants:
image

Adds Value component with different variants (optionally accepts an icon and a unit)
image


tinlake-ui/components/PoolList/index.tsx
```diff
@@ -47,6 +45,10 @@ const getDropAPY = (dropAPY: BN | null) => {
   }
 }
 
+const toNumber = (value: BN | undefined, decimals: number) => {
```
alessioscalici, line 48: Minor: is this logic used anywhere else in the app? Would it make sense to move it in a `format` collection of utility functions?

alessioscalici (approved):

PR #381, opened 2021-09-16
Add Invest button to page header on desktop

onnovisser:
image

### DIFF
--- a/tinlake-ui/components/Overview/index.tsx
+++ b/tinlake-ui/components/Overview/index.tsx
@@ -1,24 +1,74 @@
 import { Modal } from '@centrifuge/axis-modal'
-import { Box, Button, Heading, Paragraph } from 'grommet'
+import { Heading, Paragraph } from 'grommet'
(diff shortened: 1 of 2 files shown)


tinlake-ui/components/Overview/index.tsx
```diff
@@ -28,30 +78,42 @@ const Overview: React.FC<Props> = (props: Props) => {
     setModalIsOpen(false)
   }
 
+  const investButton =
```
alessioscalici, line 81: Would it make sense to extract this in a separate component / file, and import it where needed instead of injecting it through the props?

onnovisser: I would have done that if we weren't going to change the onboarding stuff. `<InvestAction />` is going to be removed next week probably, so I didn't bother.

alessioscalici: fair enough! approving!

alessioscalici (approved):

Issue #363, opened 2021-09-07
Minor fixes - Update icons / Fix spacing / Wording / 


denniswell:
- [x] Consistent use of copy icon across UI
    - Please use icon here: ![](https://storage.googleapis.com/centrifuge-hackmd/upload_fb8ddf3b334ff7a72dce95b6b776ee75.png )
https://www.figma.com/file/hXkBIpynGQX4AYCNhpskZD/Axis-Styleguide?node-id=1750%3A835
    - locations to be replaced
        - web3wallet ![](https://storage.googleapis.com/centrifuge-hackmd/upload_b1d4bfdf00013b30991cebbcfccd50aa.png)
        - Asset details ![](https://storage.googleapis.com/centrifuge-hackmd/upload_b8d1f0746b1c1eb6ff5633a54f56c2f7.png)
- [x] Remove slack icon from help menu and replace by discourse item
    - ![](https://storage.googleapis.com/centrifuge-hackmd/upload_a0f57dbb2ac763080ef485c3ce51d021.png)
    - please use this to replace: https://www.figma.com/file/hXkBIpynGQX4AYCNhpskZD/Axis-Styleguide?node-id=3026%3A3187
- [x] Fix spacing in onboarding explainer; apply "static card" style to container (white background, grey border, no drop-shadow)
![](https://storage.googleapis.com/centrifuge-hackmd/upload_59bb4972d264502fd74a73b14f577c50.png)
- [x] Move Unlock NFT button
The button should be moved into top right corner, where the 'Lock NFT' button is also positioned. Can remain secondary button. Please only display before first borrowing, otherwise hide. 
![](https://storage.googleapis.com/centrifuge-hackmd/upload_2d3ddfd9caeffc6505333b0b6cbaaa47.png)
- [x] Add risk group to asset list csv export
    - we have a feature flag `.../assets?export=true` that allows exporting the data of the asset list as csv
    - It would be great if the risk score per financing could be added to this csv export as an additional data field. This should be available as it is visible on the asset detail sheet
- [x] **Pool > Investments**: Adjust formats of epoch detail expandable:
![image](https://user-images.githubusercontent.com/23652850/132340441-d685a13e-5ead-4e71-af62-85a5c3dc9ecf.png) [Figma](https://www.figma.com/file/EWziKt1qCFdHGhWdnkrcKf/Tinlake-UI?node-id=5325%3A182)
   - Add emdash `—` as divider between epoch and status
[...]

PR #400, opened 2021-09-22
Epoch overview header

alessioscalici:
Contributes to https://github.com/centrifuge/apps/issues/363

Screenshot 2021-09-22 at 17 03 12


onnovisser (approved):
Nice!

PR #417, opened 2021-09-28
Investor onboarding

onnovisser:
Onboarding as investor view, including Onboard As Investor button on dashboard.

image

Not connected
image

Connected but not onboarded
image

The changes to the "Connect" and "Claim CFG" buttons and the explainer are also visible without the new onboarding flag enabled.


tinlake-ui/components/Onboarding/KycStep.tsx
```diff
@@ -8,34 +8,28 @@ import { StepParagraph } from './StepParagraph'
...
   state: StepProps['state']
-  onboardingData: AddressStatus | undefined
+  onboardingData: Pick<AddressStatus, 'kyc'> | undefined
   kycStatus: KycStatusLabel | 'requires-signin' | undefined
   accreditationStatus: boolean
 }
 
 // TODO: Redirect to onboard API URL that isn't pool dependant
-const placeholderPoolId = '0xc5BfCcBe24b037459922F70ADA6706638A550338'
+const placeholderPoolId = '0x560Ac248ce28972083B718778EEb0dbC2DE55740'
```
alessioscalici, line 17: Out of curiosity: why is this specific pool used as placeholder? Is there a rule, or must be any pool until the API is changed?

alessioscalici (approved):

tinlake-ui/components/Onboarding/KycStep.tsx
onnovisser: the old one was a Kovan one, but I substituted with a mainnet

tinlake-ui/components/FunnelHeader/index.tsx
```diff
@@ -45,7 +45,7 @@ export const FunnelHeader: React.FC = () => {
             getAddressLink={getAddressLink}
           />
         ) : (
-          <Button onClick={connectAccount} label="Connect" />
+          <Button primary onClick={connectAccount} label="Connect Wallet" />
```
armandomistral, line 48: @onnovisser, I know we're not consistent right now with button labels. At first I thought we should use Title case (applied here). But having a look around, I think we're actually following more sentence case (i.e. this would mean "Connect wallet"). I've applied sentence case now to all onboarding mocks, and the button component in Figma. Overall, I think sentence case is fine. Less authoritative. WDYT? Let's change to `Connect wallet`

armandomistral (approved):

tinlake-ui/components/FunnelHeader/index.tsx
onnovisser: Fine by me, will change. Same for `Onboard as investor` then?

PR #419, opened 2021-09-29
chore: add pr template

jpangelle:
### Description

This pull request adds a PR template that will auto-fill the description text area for future PR's.

[Related Github documentation](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository)

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,28 @@
+<!-- This template is a starting point for creating a pull request. Not every pull request requires thorough documentation so use your best judgement. Typically, the higher the impact, the more documentation that is warranted. Feel free to remove sections that are not applicable to the pull request or use any combination of sections that make sense. -->
+
+### Description
+
+<!-- Describe the goal of this pull request and if possible, insight on the implementation choices. -->
+
+This pull request...
+
+<!-- This will link the pull request to a Github issue. Remove if there is no corresponding Github issue. -->
+
+Closes #<issue-number>
+
+### Approvals
+
+<!-- Depending on the nature of the pull request, remove the roles that are not necessary for this review. Intricate technical changes may require two dev approvals. Reviewers should check the corresponding box after they approve. -->
+
+- [ ] Dev
+- [ ] Dev
+- [ ] Designer
+- [ ] Product
+
+### Screenshots
+
+<!-- Supporting screenshots, gifs, or videos to give reviewers an idea of how the pull request will affect the presentation of the app. -->
+
+### Impact
+
+<!-- Describe what parts of the app were affected so that reviewers can direct their focus. -->


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
...
+<!-- This will link the pull request to a Github issue. Remove if there is no corresponding Github issue. -->
+
+Closes #<issue-number>
+
+### Approvals
+
+<!-- Depending on the nature of the pull request, remove the roles that are not necessary for this review. Intricate technical changes may require two dev approvals. Reviewers should check the corresponding box after they approve. -->
+
+- [ ] Dev
+- [ ] Dev
```
alessioscalici, line 18: This is duplicated

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
...
+
+<!-- Describe the goal of this pull request and if possible, insight on the implementation choices. -->
+
+This pull request...
+
+<!-- This will link the pull request to a Github issue. Remove if there is no corresponding Github issue. -->
+
+Closes #<issue-number>
+
+### Approvals
```
alessioscalici, line 13: This is a good idea, another way to achieve a similar result (but with enforced review) would be using the code owners (although that's more related to which files are modified). Do you think we should consider it?

PULL_REQUEST_TEMPLATE.md
jpangelle: This is intentional.

> Intricate technical changes may require two dev approvals.

PULL_REQUEST_TEMPLATE.md
jpangelle: Hmm I never used that functionality. How does it work? Do you have to set up code owners on every single file?

PULL_REQUEST_TEMPLATE.md
alessioscalici: Thanks! Resolving :)

PULL_REQUEST_TEMPLATE.md
alessioscalici: Roughly, you can assign users (or teams) to file paths (directories or single files), adding a `CODEOWNERS` file to the repo. Whenever a change is made to the file included in the config file, approval of the related users/teams will be required to merge.

Here is some docs [about codeowners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)

jpangelle: Ah I see. My thinking is that this would be hard to maintain and it's not always clear who should review a PR just based on the files changed. What do you think?

alessioscalici: I think probably codeowners solve a different problem, maybe it's more useful when somebody actually owns a piece of code, meaning they built it in the first place or they know it inside and out.
About maintainability, I think it would help if too many hands get on the same code (had this kind of problem in the past) but for now I think it's unnecessary. Resolving

alessioscalici (approved):

PR #424, opened 2021-10-04
remove GasNow dependency

onnovisser:
Remove GasNow dependency for gas prices and use `maxFee` and `maxPriorityFee` instead


onboarding/api/src/utils/tx-manager.ts
```diff
@@ -62,12 +45,11 @@ class TransactionManager extends ethers.Signer {
       if (this.config.filterDuplicates && !increases && this.transactions[key]) {
         throw new Error(`Transaction ${key} already sent`)
       }
-
       this.transactions[key] = { request: transaction, resolve, reject }
       this.queue.push(key)
 
       if (this.queue.length === 1) {
-        console.log(`Processing ${key} immediately`)
+        console.log(`Procesisng ${key} immediately`)
```
offerijns, line 52:
```suggestion
        console.log(`Processing ${key} immediately`)
```

onboarding/api/src/services/pool.service.ts
```diff
@@ -20,10 +20,7 @@ export class PoolService {
   private pools: { [key: string]: Pool } = {}
 
   provider = new FastJsonRpcProvider(config.rpcUrl)
-  signer = new TransactionManager(new ethers.Wallet(config.signerPrivateKey), {
-    initialSpeed: 'fast',
-    increasedSpeed: 'rapid',
-  }).connect(this.provider)
+  signer = new TransactionManager(new ethers.Wallet(config.signerPrivateKey)).connect(this.provider)
```
offerijns, line 23: Can you set the `initialPriorityFeePerGas` here to 4 already? I think that's worth it for whitelisting, which has some time urgency, and these transactions use a small amount of gas so the impact is relatively low.

offerijns: Can you set `maxFeePerGas` to 600 for the onboarding API and 300 for the bot? A bit ridiculous, but we're now sometimes also paying this, and if gas is high for a day, we shouldn't stop closing epochs entirely.

offerijns (approved):

onboarding/api/src/services/pool.service.ts
onnovisser: updated

PR #426, opened 2021-10-04
Fix invest button for old onboarding

onnovisser:
Old invest button wasn't redirecting to pool onboarding page when pool is not upcoming. Also fixes alignment of `Onboard as Investor` button

### DIFF
--- a/tinlake-ui/components/InvestAction/index.tsx
+++ b/tinlake-ui/components/InvestAction/index.tsx
@@ -9,7 +9,7 @@ import { useInvestorOnboardingState } from '../../utils/useOnboardingState'
 import { usePool } from '../../utils/usePool'
 import { Button } from '../Button'
 import { useDebugFlags } from '../DebugFlags'
-import { PoolLink } from '../PoolLink'
+import { Center } from '../Layout'
 import { Tooltip } from '../Tooltip'
 import { FormModal, InvestmentSteps } from './styles'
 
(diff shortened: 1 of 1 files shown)


tinlake-ui/components/InvestAction/index.tsx
```diff
@@ -180,6 +146,29 @@ const InvestAction: React.FC<Props> = (props) => {
             </Box>
           )}
         </Box>
+        <Center>
+          {(props.pool as Pool)?.metadata.securitize?.issuerId ? (
+            <Button
+              primary
+              label={`Onboard as an investor`}
+              fill={false}
+              href={`https://id.securitize.io/#/authorize?registration=true&issuerId=${
```
alessioscalici, line 155: I think it would make sense to encapsulate this URL generation in its own function, trowing an error if the parameters are undefined. In this way, we avoid to redirect to securitize with `issuerId=undefined`, a bug which would be more difficult to identify and report

tinlake-ui/components/InvestAction/index.tsx
```diff
@@ -180,6 +146,29 @@ const InvestAction: React.FC<Props> = (props) => {
...
+              }&scope=info%20details%20verification&redirecturl=https://${
+                (props.pool as Pool).metadata.securitize?.slug
+              }.invest.securitize.io/%23/authorize`}
+              target="_blank"
+            />
+          ) : (
+            <Button
+              primary
+              label={`Onboard as an investor`}
+              href={`https://id.securitize.io/#/authorize?issuerId=4d11b353-a327-49ab-b45b-ae5be60697c6&scope=info%20details%20verification&registration=true&redirecturl=https://centrifuge.invest.securitize.io/#/authorize`}
```
alessioscalici, line 166: Same here, would it make sense to encapsulate this URL too?

alessioscalici (changes requested):

alessioscalici (approved):

PR #448, opened 2021-10-08
Create NFT Studio project

onnovisser:
### Description

Create NFT Studio project with barebones setup to connect Polkadot accounts.

### Approvals

- [ ] Dev


nft-studio/src/components/GlobalStyle.ts
```diff
@@ -0,0 +1,18 @@
+import { createGlobalStyle } from 'styled-components/macro'
+
+export const GlobalStyle = createGlobalStyle`
+  * {
```
alessioscalici, line 4: I would avoid creating a project-specific global style, as this could affect the external components (e.g. from Fabric) and might cause inconsistency with those components when rendered in other projects, when the global CSS is not aligned. If global CSS is needed, I would recommend providing it from a centralized source (e.g. Fabric, or  CSS reset shared by all projects)

nft-studio/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "eject": "react-scripts eject"
+  },
+  "dependencies": {
+    "@polkadot/api": "^6.2.1",
+    "@polkadot/extension-dapp": "^0.40.3",
+    "@polkadot/react-identicon": "^0.85.4",
+    "@polkadot/util-crypto": "^7.4.1",
+    "@styled-system/css": "^5.1.5",
+    "@styled-system/should-forward-prop": "^5.1.5",
+    "cra-template-typescript": "1.1.2",
```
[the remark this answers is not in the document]
onnovisser: That's what I ran, but it added it for some reason 🤷

nft-studio/src/components/Web3Provider.tsx
```diff
@@ -0,0 +1,122 @@
+import { isWeb3Injected, web3AccountsSubscribe, web3Enable, web3EnablePromise } from '@polkadot/extension-dapp'
+import { InjectedAccountWithMeta } from '@polkadot/extension-inject/types'
+import * as React from 'react'
+
+const KUSAMA_GENESIS_HASH = '0xb0a8d493285c2df73290dfb7e61f870f17b41801197a149ca93654499ea3dafe'
```
alessioscalici, line 5: Question: should this be placed in the secrets and passed through ENV variable?

onnovisser: not actually sure yet.

nft-studio/src/components/GlobalStyle.ts
alessioscalici: great!

PR #454, opened 2021-10-11
Upgrade dependencies

onnovisser:
### Description

Fix the build by upgrading some dependencies and moving Tinlake.js from `rollup` to `tsdx`

### Approvals

- [x] Dev

### Impact

Tinlake.js is being built differently, so we should check transactions still work.
Polkadot packages were updated in `tinlake-ui`, so we should check we can still connect and claim rewards.


avinashega (approved):
LGTM

PR #465, opened 2021-10-14
Fabric dark mode

onnovisser:
### Description

Adds a dark color mode to the Fabric theme

### Approvals

- [x] Dev

### Screenshots
image
image


fabric/src/theme/altairDark.ts
```diff
@@ -0,0 +1,16 @@
+import { baseTheme, FabricTheme } from './tokens/baseTheme'
+import { brandColors } from './tokens/brandColors'
+import { modes } from './tokens/modes'
+
+export const altairDark: FabricTheme = {
+  ...baseTheme,
+  colors: {
+    ...baseTheme.colors,
+    ...brandColors.altair,
+    ...modes.dark.colors,
```
alessioscalici, line 10: I think it would make sense to keep the "light" and "dark" modes in separate files, as we'll always use them separately and they should be mutually exclusive

fabric/src/theme/altairLight.ts
```diff
@@ -0,0 +1,16 @@
+import { baseTheme, FabricTheme } from './tokens/baseTheme'
+import { brandColors } from './tokens/brandColors'
+import { modes } from './tokens/modes'
```
alessioscalici, line 3: Same here, i would suggest to split the modes in 2 files and `import lightMode from './tokens/lightMode'`

fabric/src/theme/index.ts
```diff
@@ -1,20 +1,12 @@
...
-  space,
-  colors,
-  sizes: {
-    container: 1152,
-  },
-  fonts: {
-    standard:
-      "AvenirNextLTW01, 'Avenir Next', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'",
-  },
+export const themes = {
```
alessioscalici, line 6: I suggest to avoid exporting all the themes in a single object: if this file is used, all the included themes will be loaded in the application.

The themes would still be importable like this: `import altairDark from '@centrifuge/fabric/theme/altairDark'`. It's a bit more verbose, but it only happens once per application

fabric/src/theme/tokens/modes.ts
```diff
@@ -0,0 +1,102 @@
...
+  okDark: okLight,
+  warningLight: warningDark,
+  warningPrimary,
+  warningDark: warningLight,
+  criticalLight: criticalDark,
+  criticalPrimary,
+  criticalDark: criticalLight,
+}
+
+export const modes = {
```
alessioscalici, line 92: This file could be split in `lightMode.ts` and 'darkMode.ts', as they are mutually exclusive and can be imported separately, avoiding loading more config than necessary

fabric/src/theme/tokens/breakpoints.ts
```diff
@@ -4,4 +4,4 @@ breakpoints.medium = breakpoints[1]
 breakpoints.large = breakpoints[2]
 breakpoints.xlarge = breakpoints[3]
```
alessioscalici, line 5: I suggest to rename the breakpoints as per design system (S, M, L and XL instead of large, xlarge etc)

fabric/src/theme/tokens/brandColors.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const centrifuge = {
+  brand: centrifugeBlue,
+}
+
+const altair = {
+  brand: altairYellow,
+}
+
+export const brandColors = {
```
alessioscalici, line 12: I suggest to split this file in 2: `brandAltair.ts` and `brandCentrifuge.ts`

The reason is, we don't need Centrifuge colors in Altair and vice-versa, so we might want to allow importing only one or the other (mutual exclusivity)

fabric/src/theme/tokens/baseColors.ts
```diff
@@ -0,0 +1,8 @@
+const baseColors = {
+  centrifugeBlue: '#2762FF',
+  centrifugeOrange: '#FCBA59',
+  altairYellow: '#FAB961',
```
alessioscalici, line 4: These are brand related colors, I would add them to the related brand colors files

alessioscalici (changes requested):

fabric/src/theme/tokens/modes.ts
onnovisser: done

fabric/src/theme/index.ts
onnovisser: done

fabric/src/components/Card/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Box, BoxProps } from '../Box'
+
+type Props = {
+  interactive?: boolean
+}
+
+export type CardProps = Props &
+  Omit<BoxProps, 'border' | 'borderColor' | 'borderWidth' | 'borderStyle' | 'bg' | 'backgroundColor'>
+
+export const Card = styled(Box)<Props>((props) =>
```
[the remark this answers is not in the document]
alessioscalici: Oh, didn't see it! Thanks!

PR #469, opened 2021-10-18
implement generic kyc

avinashega:
### Description

This pull request implements the getStatusForAddress endpoint to get KYC status for an address.

Closes #<issue-number>

### Approvals

- [x] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product


onboarding/api/src/services/docusign.service.ts
```diff
@@ -42,6 +42,8 @@ export class DocusignService {
       status: 'sent',
     }
 
+    console.log(envelopeDefinition)
```
alessioscalici, line 45: Is this a debug leftover? Should it be removed?

PR #470, opened 2021-10-19
Fabric: Button component

onnovisser:
### Description

Updates the Fabric button component. Separates the Visual from the functional part of the button.

### Approvals

- [ ] Dev

### Screenshots

image
image

### Impact

Fabric


fabric/src/components/Button/VisualButton.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const disabled = disabledProp || loading
+
+  let fg,
+    bg,
+    fgHover,
+    bgHover,
+    borderWidth = 0
+  if (variant === 'contained') {
+    fg = 'backgroundPrimary'
+    bg = disabled ? 'textDisabled' : 'textPrimary'
```
alessioscalici, line 127: Here `fg` and `bg` look swapped, bg probably shouldn't get `text*` colors

fabric/src/components/Button/VisualButton.tsx
```diff
@@ -0,0 +1,164 @@
...
+}) => {
+  const iconSize = variant !== 'text' || small ? 'iconSmall' : 'iconMedium'
+  const disabled = disabledProp || loading
+
+  let fg,
+    bg,
+    fgHover,
+    bgHover,
+    borderWidth = 0
+  if (variant === 'contained') {
```
alessioscalici, line 125: As this logic is to decide the CSS values of the style button, I think it belongs to the `StyledButton` component:

I would pass along only the `variant`, `disabled` and `small` properties (without the $ prefix), and do something like:

```typescript
({ variant, disabled, small }) => {
  const fgColor = variant === 'contained' ? 'textPrimary' : disabled ? 'textDisabled' : 'textPrimary';

    const fgHover = variant === 'contained' ? fgColor : 'brand'
    return css({
      color: fgColor,
      backgroundColor: props.$bg,
      borderColor: fgColor,
      borderWidth: variant === 'outlined' ? 1 : 0,
      pointerEvents: disabled ? 'none' : 'initial',
      height: small ? 32 : 40,

      '&:hover, &:active': {
        color: fgHover,
        borderColor: fgHover,
        backgroundColor: variant === 'contained' ? 'brand' : 'transparent',
      },

      '&:active': {
        boxShadow: variant !== 'text' ? 'buttonFocused' : 'none',
      },
    })
}
```

fabric/src/components/Button/VisualButton.tsx
alessioscalici: I see, thanks! Or also leaving a comment to explain, could also be good, otherwise it looks like a mistake :D

PR #495, opened 2021-10-29
Nft studio Pin image file and metadata

alessioscalici:
### Description

This pull request adds a basic form that allows to pin an image file and a json metadata to IPFS using Pinata

### Approvals

- [ ] Dev

### Screenshots
Screenshot 2021-10-29 at 17 51 15


nft-studio/lambda/pinFile.js
```diff
@@ -0,0 +1,111 @@
...
+  return axios.post(`${PINATA_BASE_URL}/pinning/pinFileToIPFS`, data, {
+    maxBodyLength: 'Infinity', // this is needed to prevent axios from erroring out with large files
+    headers: {
+      'Content-Type': `multipart/form-data; boundary=${data._boundary}`,
+      ...PINATA_AUTH_HEADERS,
+    },
+  })
+}
+
+const ipfsHashToURI = (hash) => `https://ipfs.io/ipfs/${hash}`
```
onnovisser, line 51: I think the idiomatic way to format an IPFS url is `ipfs://ipfs/{hash}`. That way any frontend can easily use it's own IPFS gateway of choice.

nft-studio/src/pages/MintNFT.tsx
```diff
@@ -1,6 +1,96 @@
...
+import { getFileDataURI } from '../utils/getFileDataURI'
 
 export const MintNFTPage: React.FC = () => {
-  return <PageContainer variant="overlay"></PageContainer>
+  const [nftName, setNftName] = useState('')
+  const [nftDescription, setNftDescription] = useState('')
+  const [fileDataUri, setFileDataUri] = useState('')
+  const [fileName, setFileName] = useState('')
+
+  const isFormValid = useMemo(() => nftName && nftDescription && fileDataUri, [nftName, nftDescription, fileDataUri])
```
onnovisser, line 14: small nit: useMemo is a bit overkill here. I think `const isFormValid = nftName && nftDescription && fileDataUri` is fine. Using useMemo is actually costlier in this case, because it's gonna check the dependency array each render.

onnovisser (approved):
Nice!

nft-studio/src/pages/MintNFT.tsx
alessioscalici: removed, thanks!

nft-studio/lambda/pinFile.js
alessioscalici: changed :)

PR #500, opened 2021-11-08
NFT Studio: Account dropdown

onnovisser:
### Description

Adds the account dropdown to the NavBar and a couple supporting Components: A Menu (should be moved to Fabric), Positioner and Popover.

### Approvals

- [ ] Dev

### Screenshots

image
image

### Impact

NFT Studio


fabric/src/components/GlobalStyle/index.ts
```diff
@@ -25,4 +30,11 @@ export const GlobalStyle = createGlobalStyle`
   *:focus {
     outline: none;
   }
+
+  button {
```
alessioscalici, line 34: I would avoid any unnecessary global style in Fabric. While in the simple case of NFT studio it doesn't bring any issue, it could interfere with styles from 3rd party components, or even from the application implementing Fabric. i would rather let each component handle the margin, padding, appearance etc. properties locally to the component, for proper CSS isolation

nft-studio/src/components/PopoverMenu.tsx
```diff
@@ -0,0 +1,53 @@
+import { Card, Shelf, Stack, Text } from '@centrifuge/fabric'
+import * as React from 'react'
+import { ResponsiveValue } from 'styled-system'
+
+type Props = {
+  children: ({ menuTriggerProps }: any) => React.ReactNode
+}
+
+export const PopoverMenu: React.FC<Props> = ({}) => {
```
alessioscalici, line 9: Since no arguments are used, this should probably be `()`

onnovisser: oops, didn't mean to even commit that file. deleted

alessioscalici (approved):

PR #513, opened 2021-11-11
NFT Studio: get available class id and create collection metadata

alessioscalici:
### Description

This pull request adds logic to fetch an available class id and to pin collection metadata on IPFS

### Approvals

- [ ] Dev

### Impact

NFT Studio collection creation


onnovisser (approved):
Nice!

PR #524, opened 2021-11-17
NFT Studio: Mint NFT transaction and error messages

onnovisser:
### Description

* Adds transaction for minting
* Displays nickname instead of address if set
* Adds error messages if balance is too low when minting/creating collection/transferring
* Fix check if user is owner of collection and can mint
* Reset forms on success

### Approvals

- [ ] Dev

### Screenshots

image

image

### Impact

NFT Studio

### DIFF
--- a/nft-studio/src/components/CreateCollectionDialog.tsx
+++ b/nft-studio/src/components/CreateCollectionDialog.tsx
@@ -11,6 +11,9 @@ import { useCreateTransaction } from '../utils/useCreateTransaction'
 import { TextArea } from './TextArea'
 import { TextInput } from './TextInput'
 
+// TODO: replace with better fee estimate
+const CREATE_FEE_ESTIMATE = 2
+
 export const CreateCollectionDialog: React.FC<{ open: boolean; onClose: () => void }> = ({ open, onClose }) => {
   const queryClient = useQueryClient()
   const { selectedAccount } = useWeb3()
@@ -54,7 +57,7 @@ export const CreateCollectionDialog: React.FC<{ open: boolean; onClose: () => vo
     }
   }, [queryClient, lastCreatedTransaction?.status])
 
-  const balanceLow = !balance || balance < 2
+  const balanceLow = !balance || balance < CREATE_FEE_ESTIMATE
 
   const disabled = !isConnected || !name || balanceLow
 
--- a/nft-studio/src/components/CreateCollectionDialog.tsx
+++ b/nft-studio/src/components/CreateCollectionDialog.tsx
@@ -1,5 +1,4 @@
-import { Button, Stack, Text } from '@centrifuge/fabric'
-import { ApiPromise } from '@polkadot/api'
+import { Button, Shelf, Stack, Text } from '@centrifuge/fabric'
 import * as React from 'react'
 import { useQueryClient } from 'react-query'
 import { ButtonGroup } from '../components/ButtonGroup'
@@ -7,6 +6,7 @@ import { Dialog } from '../components/Dialog'
 import { useWeb3 } from '../components/Web3Provider'
 import { createCollectionMetadata } from '../utils/createCollectionMetadata'
 import { getAvailableClassId } from '../utils/getAvailableClassId'
+import { useBalance } from '../utils/useBalance'
 import { useCreateTransaction } from '../utils/useCreateTransaction'
 import { TextArea } from './TextArea'
 import { TextInput } from './TextInput'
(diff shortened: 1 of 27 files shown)


nft-studio/src/components/CreateCollectionDialog.tsx
```diff
@@ -52,9 +54,13 @@ export const CreateCollectionDialog: React.FC<{ open: boolean; onClose: () => vo
     }
   }, [queryClient, lastCreatedTransaction?.status])
 
+  const balanceLow = !balance || balance < 2
```
alessioscalici, line 57: Where does the number 2 come from? Should we extrapolate it into a config, to allow reuse?

nft-studio/src/components/Identity.tsx
```diff
@@ -0,0 +1,39 @@
...
+      title={addr}
+      style={{ cursor: clickToCopy ? 'copy' : undefined }}
+      onClick={clickToCopy ? () => copyToClipboard(addr) : undefined}
+    >
+      {identity?.display || truncateAddress(address)}
+    </Text>
+  )
+}
+
+function copyToClipboard(value: string) {
```
alessioscalici, line 27: I think this is a good candidate to be moved to its own file, as the clipboard functionality is likely to be required elsewhere too

nft-studio/src/components/TransferDialog.tsx
```diff
@@ -61,9 +64,13 @@ export const TransferDialog: React.FC<Props> = ({ open, onClose, collectionId, n
 
   const error = getError()
 
+  const balanceLow = !balance || balance < 1 // TODO: replace with transaction fee estimate
```
alessioscalici, line 67: Also this number 1, I would suggest to move it to a config/constant with an explicative name

nft-studio/src/pages/MintNFT.tsx
```diff
@@ -103,12 +137,19 @@ export const MintNFTPage: React.FC = () => {
...
+                <Button disabled={disabled} type="submit" loading={isMinting}>
                   Mint
                 </Button>
                 <RouterLinkButton to={`/collection/${collectionId}`} variant="outlined">
                   Cancel
                 </RouterLinkButton>
+                {(balanceLow || !canMint) && (
+                  <Text variant="label1" color="criticalForeground">
+                    {!canMint
+                      ? `You're not the owner of the collection`
```
alessioscalici, line 149: Note for the design: I would suggest that the "mint" button is hidden when:

- The user wallet is not connected
- The user is not the owner of the collection

This would allow to avoid a click + error scenario.

Alternatively to hiding the button, we could disable it and add a small informative text close to it, like "Connect your wallet to mint new NFTs".

@armandomistral what do you think?

onnovisser: The Mint button is already hidden in that case, but a user could accidentally switch account when already on the Mint NFT page, hence the extra check here

nft-studio/src/utils/web3.ts
```diff
@@ -3,8 +3,8 @@ import { addressEq, encodeAddress } from '@polkadot/util-crypto'
 
 export function truncateAddress(address: string) {
   const encodedAddress = encodeAddress(address, 2)
-  const first8 = encodedAddress.slice(0, 8)
-  const last3 = encodedAddress.slice(-3)
+  const first8 = encodedAddress.slice(0, 6)
```
alessioscalici, line 6: This names are now misleading (first8 contains 6 character now) I would suggest to rename the variables as `head` and `tail`, or any other name that doesn't imply the length of the string

nft-studio/src/components/TransferDialog.tsx
onnovisser: It should be using the `useTransactionFeeEstimate` but the pallet for getting the fee info isn't available yet. So hence a rough estimate

nft-studio/src/utils/web3.ts
onnovisser: lol, missed that

nft-studio/src/components/CreateCollectionDialog.tsx
alessioscalici: When we'll have an exact estimate, we'll need to go through the components to look to a number 2 to replace! I still think that moving it to a constant with an explicit name (and maybe a TODO / FIXME comment) would make that task much easier

nft-studio/src/pages/MintNFT.tsx
alessioscalici: Cool, thanks! As nothing said then :)

nft-studio/src/components/CreateCollectionDialog.tsx
onnovisser: true, will do that

alessioscalici (approved):

PR #539, opened 2021-12-02
Setup centrifuge-js

onnovisser:
### Description

Initial setup of the `centrifuge-js` package. There's a number of API design decisions that could use some discussion. Here are a couple, but please chime in with more thoughts, ideas, etc.

#### Passing config
 Currently takes a `network` option, which can be `altair` or `centrifuge`. It then has some default config to get the right relaychain/parachain urls. Alternative could be just passing the urls as config, and Centjs not having any knowledge of which network it is.
#### Passing signer
Currently has a `connect` function to add a signer which returns a new Centrifuge instance with that signer. Alternative could be just having a `setSigner` function which sets some variables. but would make the instance more stateful.
#### Transactions and payment info
The example transaction takes two arguments `args` and optional `options`. It returns a promise that resolves when the transaction is successful. passing the `paymentInfo` option doesn't call the extrinsic but instead returns the gas fee estimate. 
```ts
const cent = new Centrifuge()
const paymentInfo = await cent.nfts.transferNft([1, 2, '5G6D...WxtR'], { paymentInfo: '5CHx...Ww3f' })
await cent.nfts.transferNft([1, 2, '5G6D...WxtR'])
```
Alternative could be that the function doesn't immediately call the extrinsic, but instead returns an object like 
```ts
{
  submit(),
  paymentInfo(address: string),
}
```

```ts
const cent = new Centrifuge()
const tx = await cent.nfts.transferNft([1, 2, '5G6D...WxtR'])
const paymentInfo = await tx.paymentInfo('5CHx...Ww3f')
await tx.submit()
```

### Approvals

- [ ] Dev
- [ ] Dev


centrifuge-js/src/CentrifugeBase.ts
```diff
@@ -0,0 +1,97 @@
...
+  altairWsUrl: string
+  polkadotWsUrl: string
+  kusamaWsUrl: string
+  signer?: Signer
+  signingAddress?: AddressOrPair
+}
+
+type UserProvidedConfig = Partial<Config>
+
+const defaultConfig: Config = {
```
offerijns, line 19: Yeah this is great, having a default config.

centrifuge-js/src/modules/nfts.ts
```diff
@@ -0,0 +1,54 @@
...
+            id,
+            collectionId,
+            owner: instance.owner,
+            metadataUri: (metas[i]?.toHuman() as any)?.data,
+          }
+          return nft
+        })
+        return mapped
+      },
+      transferNft: async (args: Parameters<ApiPromise['tx']['uniques']['transfer']>, options?: TransactionOptions) => {
```
offerijns, line 47: Maybe it would be better to always create custom args interfaces, rather than copy those coming from our metadata. I think that would make migrations easier?

centrifuge-js/src/modules/nfts.ts
```diff
@@ -0,0 +1,54 @@
...
+export type NFT = Instance & {
+  id: string
+  collectionId: string
+  metadataUri?: string
+}
+
+export function WithNfts<Base extends CentrifugeBaseConstructorLike>(base: Base) {
+  return class extends base {
+    nfts = {
+      getAccountNfts: async (address: string) => {
```
offerijns, line 21: Should we be consistent and always use an args object, as @alessioscalici suggested?

nft-studio/package.json
```diff
@@ -16,6 +17,7 @@
     "eject": "react-scripts eject"
   },
   "dependencies": {
+    "@centrifuge/centrifuge-js": "workspace:*",
```
offerijns, line 20: Hm... this feels a bit redundant. How about just `@centrifuge-js`, or `@centrifuge/js`, or `@centrifuge/api`, or `@centrifuge/client`? Just some other ideas.

offerijns: Although actually Stripe uses the same naming: https://www.npmjs.com/package/@stripe/stripe-js

alessioscalici: In this case, even though I don't like the redundancy, I think `@centrifuge/centrifuge-js` makes sense, because the `@centrifuge` is a namespace we use for everything, while the package name (and the directory in the repo) are explanatory of the content. So for consistency, I think it makes sense to keep it

centrifuge-js/src/modules/nfts.ts
alessioscalici: If we do that, I think we should also pay extra care about the arguments naming. We could think to standardise it in a flexible but **consistent** way: for example, we could decide that every method needing the current connected address would have the e.g. `address`. If another method needs 2 addresses (e.g. source and destination) should we keep the name `address` for the source (for consistency), and use another name for the destination e.g. `toAddress` (or `destAddress`, `destinationAddress` etc). Of course this is only an example.

WDYT?

centrifuge-js/src/modules/nfts.ts
onnovisser: yeah, makes sense

centrifuge-js/src/modules/nfts.ts
onnovisser: it could just be an args array similar to transactions, then the names of the args would be purely descriptive

centrifuge-js/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "lint": "tsdx lint",
+    "prepare": "tsdx build"
+  },
+  "dependencies": {
+    "@polkadot/api": "^6.2.1",
+    "@polkadot/util-crypto": "^7.4.1"
+  },
+  "devDependencies": {
+    "tsdx": "^0.14.1",
+    "tslib": "^2.3.1",
```
jpangelle, line 27: Haven't seen these two libraries before. Curious on the differences.

onnovisser: [tsdx](https://tsdx.io/) is a cli for easily building typescript packages, basically a wrapper around RollupJS. the tslib package it added automatically when running the cli to set up the package. Not sure what it's use is actually. I can try removing it, see if it breaks anything.

centrifuge-js/src/modules/nfts.ts
```diff
@@ -0,0 +1,199 @@
...
+      const id = String(getRandomUint())
+      const res = await api.query.uniques.class(id)
+      if (res.toJSON() === null) {
+        return id
+      }
+    }
+    throw new Error(`Could not find an available collection ID in ${MAX_ATTEMPTS} attempts`)
+  }
+
+  async function getAvailableNftId(collectionId: string) {
```
onnovisser, line 177: Thoughts on if this should also use an args array? I don't really see this as a query, but more a utility.

centrifuge-js/src/modules/nfts.ts
```diff
@@ -0,0 +1,199 @@
...
+}
+
+export type Collection = Class & {
+  id: string
+  metadataUri?: string
+}
+
+const MAX_ATTEMPTS = 10
+
+const getRandomUint = () => (Math.random() * (2 ** 53 - 1)) >>> 0
```
offerijns, line 39: Maybe we should move these into a utils folder?

centrifuge-js/src/modules/nfts.ts
```diff
@@ -0,0 +1,199 @@
...
+const formatClassKey = (keys: StorageKey<[u32]>) => (keys.toHuman() as string[])[0].replace(/\D/g, '')
+const formatInstanceKey = (keys: StorageKey<[u32, u32]>) => (keys.toHuman() as string[])[1].replace(/\D/g, '')
+
+export function getNftsModule(inst: CentrifugeBase) {
+  async function getCollections() {
```
offerijns, line 44: Maybe I'm missing something, but shouldn't we be specifying return types for these methods?

onnovisser: it infers it from the return type
image

centrifuge-js/src/modules/nfts.ts
onnovisser: yea might as well

onnovisser:
replaced tsdx with esbuild

alessioscalici:
@onnovisser is this still relevant? As we have the `centrifuge-js` package in place, should this PR be closed?

PR #544, opened 2021-12-08
Fabric: Update and fix Svgr

onnovisser:
### Description

`npx` was using the latest version of Svgr, regardless of the version in the `package.json` . There was a new major version released of Svgr, which broke the build. This updates Svgr and changes it to use the installed version instead.

### Approvals

- [ ] Dev

### Impact

Fabric


alessioscalici (approved):

alessioscalici:
Nice!

PR #553, opened 2021-12-16
fix: missed investors query

offerijns:


EstherCentrifuge:
Can you please add a description and add it to the correct epic @offerijns 

do we need it for first pool? -.>V1
do we need it before migration? -> v2
If no -> filler

EstherCentrifuge:
This is "open" and 8+months old can you updaet please> @offerijns

PR #600, opened 2022-01-24
Send email via sendgrid on whitelist

avinashega:
### Description

This pull request triggers the invest now email when a user signs the docusign agreement.

Closes #<issue-number>

### Approvals

- [ ] Dev

### Impact

Replace third-party emails with our own emails to have better control over the content.


onboarding/api/src/config.ts
```diff
@@ -84,6 +89,11 @@ const config: Config = {
   rwaMarket: {
     permissionManagerContractAddress: process.env.RWA_MARKET_PERMISSION_MANAGER_CONTRACT_ADDRESS,
   },
+  sendgrid: {
+    templateId: `d-a0e7c096a0cf43868de275751e57fdf5`,
```
avinashega, line 93: Wondering if there is a better place to define this!

onboarding/api/src/utils/mailer.ts
```diff
@@ -0,0 +1,35 @@
+import config from '../config'
+import { User } from '../repos/user.repo'
+
+export class Mailer {
+  async sendEmail(user: User, pool: any) {
+    const response = await fetch(config.sendgrid.apiUrl, {
+      body: JSON.stringify({
+        from: {
+          name: 'Tinlake pools',
+          email: `issuer+${pool.profile.issuer.name}@tinlake.com`,
```
offerijns, line 10: I think the issuer name can contain spaces, so this should be lowercased and spaces should be swapped for `-`s.

onboarding/api/src/utils/mailer.ts
```diff
@@ -0,0 +1,35 @@
...
+        },
+        personalizations: [
+          {
+            to: [user.email],
+            dynamic_template_data: {
+              investorName: user.fullName,
+              poolName: pool.metadata.name,
+              token: 'DROP',
```
offerijns, line 18: This probably needs to be the full token name, and is dependent on the tranche being invested in. E.g. `${pool.metadata.name} ${tranche}`

onboarding/api/src/utils/mailer.ts
```diff
@@ -0,0 +1,35 @@
+import config from '../config'
+import { User } from '../repos/user.repo'
+
+export class Mailer {
+  async sendEmail(user: User, pool: any) {
+    const response = await fetch(config.sendgrid.apiUrl, {
+      body: JSON.stringify({
+        from: {
+          name: 'Tinlake pools',
```
offerijns, line 9: This should probably be `pool.profile.issuer.name`.

onboarding/api/src/utils/mailer.ts
```diff
@@ -0,0 +1,35 @@
+import config from '../config'
+import { User } from '../repos/user.repo'
+
+export class Mailer {
+  async sendEmail(user: User, pool: any) {
```
offerijns, line 5: We should already adapt this for multiple types of emails in the future. E.g. rename this to `sendWhitelistedConfirmation()`. And then some of the sendgrid logic can be refactored into a `sendEmail` method, which accepts `templateId`, `templateData`, `to`, etc as parameters.

onboarding/api/src/config.ts
offerijns: We'll need different template ids for different emails. I think it's okay to keep this in the config file hardcoded for now.

offerijns (changes requested):
Good start 👍

Issue #624, opened 2022-02-22
POCC MVP - Display important values with 4 digits


denniswell:
Even if it doesn't look pretty ;-). These values are:

Asset view:
Current debt

Token view:
![image.png](https://images.zenhubusercontent.com/5c922180cf9c5c1a350a748b/7c325bc1-f96d-4213-8708-0bef3987bbfe)
All values please ;-)

Managed pools view:
![image.png](https://images.zenhubusercontent.com/5c922180cf9c5c1a350a748b/f30e5f3a-4502-4a83-9f90-15c97c347dfc)
Pool Value 
Asset Value 
Reserve

In the tranche components
Tranche Value 
Token price (when added)

Issue #623, opened 2022-02-22
POCC MVP - Add discount rate to asset details


denniswell:
Please add discount rate of asset here:

![image.png](https://images.zenhubusercontent.com/5c922180cf9c5c1a350a748b/d07815f2-bef7-4627-88b9-bb0b6b0e6e0f)

PR #625, opened 2022-02-22
POCC: Add decimals

### LINKED ISSUE #622: POCC MVP - Update to token component
THe token component currently looks like this:

The following changes would be helpful to make testing easier:
Remove Tranche `Reserve` 
Add `Token price` (decimal, 4 digits)
Add `Currently locked investment` (decimal, 0 digits, 0 if nothing locked)
[Nice to have] Add `Amount to be collected` (decimal, 0 digits, 0 if nothing to be collected)

It would be great if the tranche value would have 4 digits as well (see other ticket)

onnovisser:
### Description

* Adds more decimals to certain displayed values
* Adds token price and invested amount to tranche card

Closes #622, #623, #624

### Approvals

- [x] Dev

### Impact

POCC

### DIFF
--- a/centrifuge-js/src/utils/index.ts
+++ b/centrifuge-js/src/utils/index.ts
@@ -81,10 +81,10 @@ export function addThousandsSeparators(x: string | BN | number) {
   return parts.join('.')
 }
 
-export function formatCurrencyAmount(bn: BN | string | undefined, currency?: string) {
+export function formatCurrencyAmount(bn: BN | string | undefined, currency?: string, precise?: boolean) {
   const currencyStr = currency || 'Usd'
   if (!bn) return ''
-  return `${addThousandsSeparators(toPrecision(baseToDisplay(new BN(bn), 18), 0))} ${currencyStr}`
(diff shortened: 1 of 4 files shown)


centrifuge-js/src/utils/index.ts
```diff
@@ -81,10 +81,10 @@ export function addThousandsSeparators(x: string | BN | number) {
   return parts.join('.')
 }
 
-export function formatCurrencyAmount(bn: BN | string | undefined, currency?: string) {
+export function formatCurrencyAmount(bn: BN | string | undefined, currency?: string, precise?: boolean) {
```
alessioscalici, line 84: I wonder if it was more flexible to pass a number (the precision) instead of a boolean?

onnovisser: Maybe. I feel like for formatting currency amounts there shouldn't be too many options. Flexibility can still be achieved by using the `toPrecision` and such themselves. Not that I'm a huge fan of this function as it is. I think we should at some point change this to use `Intl.NumberFormat` or `Number.toLocaleString` instead, and then reconsider the options.

alessioscalici: fair enough 😄

alessioscalici (approved):

PR #626, opened 2022-02-23
Polkadot typegen

alessioscalici:
### Description

This pull request adds automatic Typescript type generation for Polkadot API to `nft-studio`

### Approvals

- [ ] Dev
- [ ] Dev


offerijns (approved):
Looks great to me! At some point we'll need to make it a bit more complex, as we'll have different metadata per runtime (and maybe per block if we want to handle migrations). But we don't need to needlessly overcomplicate this now.

centrifuge-js/src/utils/web3.ts
```diff
@@ -1,13 +1,16 @@
 import { ApiPromise, WsProvider } from '@polkadot/api'
 import { RegistryTypes } from '@polkadot/types/types'
+import * as definitions from '../interfaces/definitions'
 
 const cached: { [key: string]: Promise<ApiPromise> } = {}
-export function getPolkadotApi(wsUrl: string, types?: RegistryTypes) {
+export function getPolkadotApi(wsUrl: string, regTypes?: RegistryTypes) {
+  const pdTypes = Object.values(definitions).reduce((res, { types }): object => ({ ...res, ...types }), {})
```
onnovisser, line 7: These types should probably only get added when we're getting the parachain api. Also maybe an idea to return the cached api if available before doing this work.

alessioscalici: Sorry, not sure I follow... aren't we doing this in this method already?

onnovisser: not sure how big those definitions are but spreading the objects inside of a loop could be a bit slow. We could return `cached[wsUrl]` before the `reduce`

onnovisser: but the main point was that this function gets called to get the parachain api, but also to get the relay chain api, which doesn't need the types

alessioscalici: I will put the cache check and early return before the type loop, good point!

Would it make sense to split the 2 functions `getRelayChainApi` and `getParachainApi` at `centrifuge-js` level? WDYT?

EstherCentrifuge:
This is deployed but "open" and 6+months old can you updaet please> @offerijns

onnovisser:
Picking this up later

Issue #51, opened 2020-12-14
Re-implement URL redirects 


offerijns:
- redirects /[ID] to /pool/[ID]
- redirects /pool/[ID] to /pool/[ID]/[alias]
- redirects /pool/[ID]/* to /pool/[ID]/[alias]/*
- redirects /pool/[alias] to /pool/[ID]/[alias]

PR #630, opened 2022-02-25
Rewrite redirects

alessioscalici:
### Description

This pull request redirects the user to the pool page when they use an URL with the following formats:

- `/[ID]` --> `/pool/[ID]/[slug]`
- `/pool/[ID]` --> `/pool/[ID]/[slug]`
- `/[slug]` --> `/pool/[ID]/[slug]`

Closes [#51](https://github.com/centrifuge/apps/issues/51)

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Product

### Impact

Tinlake UI

### DIFF
--- a/tinlake-ui/package.json
+++ b/tinlake-ui/package.json
@@ -15,13 +15,14 @@
     "start": "yarn build:deps && run-p -l dev:next dev:lambda",
     "serve": "next start",
     "test": "jest",
-    "export": "yarn build && next export",
+    "export": "yarn build && next export && yarn build:redirects",
     "lint": "eslint -c .eslintrc.js '*/**/*.{js,ts,tsx}'",
     "lint-fix": "eslint -c .eslintrc.js '*/**/*.{js,ts,tsx}' --fix",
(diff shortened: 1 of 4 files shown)


offerijns (approved):
That's a neat solution! We should do some more UI testing (basically clicking around the app and refreshing pages, so make sure nothing is broken), but the code implementation looks great.

onnovisser (approved):

alessioscalici:
@offerijns thanks! I have tested this extensively again this morning, all the combinations with different pools, all the tested links worked fine. Only thing it might be worth to add in my opinion, is to redirect the subpaths in all cases:

In the ticket issue, redirects `/pool/[ID]/*` to `/pool/[ID]/[alias]/*` means for example /`pool/[ID]/assets` to `/pool/[ID]/[alias]/assets`, but this is not indicated for the other options. Is there a reason?

Do you think we should redirect also `/pool/[alias]/*` and `./[ID]/*`?

Also, what about `/[alias]/*` (without the `/pool` prefix)?

PR #678, opened 2022-03-04
Redirect to collection page after creation

alessioscalici:
### Description

This pull request:
- redirect to collection page after creation
- disables buttons and fields while the transaction is pending

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product

### Screenshots

Screenshot 2022-03-04 at 14 46 04


nft-studio/src/components/CreateCollectionDialog.tsx
```diff
@@ -33,9 +37,12 @@ export const CreateCollectionDialog: React.FC<{ open: boolean; onClose: () => vo
     reset: resetLastTransaction,
     isLoading: transactionIsPending,
   } = useCentrifugeTransaction('Create collection', (cent) => cent.nfts.createCollection, {
-    onSuccess: () => {
+    onSuccess: ([collectionId]) => {
       queryClient.invalidateQueries('collections')
       queryClient.invalidateQueries('balance')
+      if (open && isPageUnchanged()) {
```
onnovisser, line 43: I think doing the redirect via state would be better here. The user could have closed the dialog and could be looking at other collections or have the dialog closed and reopened to create another collection. I think redirecting in those cases would be bad UX. Doing it via state here would make sure it only redirects if they still have the same dialog open.

alessioscalici: good point, I'll change the implementation!

alessioscalici: Done!

onnovisser (approved):

PR #688, opened 2022-03-08
swap subgraph to tax-reporting

AStox:
### Description

This pull request swaps out the subgraph to the latest version `tax-reporting` which is identical to the previous subgraph `investor-transactions-and-split-rewards` but includes the two missing coordinator contracts.

### Approvals

- [ ] Dev


jpangelle (approved):
LGTM!

PR #695, opened 2022-03-09
Send email via sendgrid on whitelist

avinashega:
### Description

This pull request triggers the invest now email when a user signs the docusign agreement.

Closes #<issue-number>

### Approvals

- [ ] Dev


.github/workflows/onboard-api-staging.yml
```diff
@@ -0,0 +1,64 @@
+name: 'onboarding-api-staging'
+on: 
+  pull_request:
+    paths:
+    - 'onboarding/api/**'
+    - '.github/workflows/onboard-api-staging.yml'
```
jpangelle, line 6: How stable do we intend this "staging" environment to be? Perhaps it shouldn't run on `pull_request`, but instead on merge to `main` or some other pattern match like we do for [`tinlake-ui`](https://github.com/centrifuge/apps/blob/main/.github/workflows/tinlake-ui-deploy-to-staging.yml#L5) and [`nft-studio`](https://github.com/centrifuge/apps/blob/main/.github/workflows/nft-studio-deploy-staging.yml#L5).

Or like `nft-studio` and `tinlake-ui`, we can have a workflow for each environment (pull request, dev, staging, and prod).

jpangelle (approved):
lgtm! just curious if we can make the release process more granular but idk if that is out of the scope of this pr

.github/workflows/onboard-api-staging.yml
avinashega: The intention was to be able to test the feature before merging to the main branch. We only have two instances of onboarding-api running at the moment - prod & staging

PR #717, opened 2022-03-11
NFT Studio: Add collection logo

onnovisser:
### Description

Adds possibility for uploading logo in Create Collection Dialog

### Approvals

- [ ] Dev
- [ ] Designer

### Screenshots

image

image


nft-studio/src/pages/Collection.tsx
```diff
@@ -62,16 +66,41 @@ const Collection: React.FC = () => {
...
+            justifyContent="center"
+            borderRadius="50%"
+          >
+            <LogoAltair width="100px" />
+          </Shelf>
+        )}
         <Stack alignItems="center" gap="4px">
           <Text variant="heading1" fontSize="36px" textAlign="center" style={{ wordBreak: 'break-word' }}>
-            {metadata?.name || ''}
+            {metadata?.name || 'Unnamed collection'}
```
alessioscalici, line 98: Not in this PR domain, but I would suggest to use loaders / spinners / skeleton templates or similar, instead of using "Unnamed X". I think we should use "Unnamed X" only after loading, for things without a name (which shouldn't happen for collections created using our UI)

alessioscalici (approved):

nft-studio/src/pages/Collection.tsx
onnovisser: Yes, I was thinking the same thing.

PR #740, opened 2022-03-17
NFT playground: Add terms

onnovisser:
### Description

Add terms to Create Collection dialog

### Approvals

- [x] Dev
- [ ] Designer
- [ ] Product

### Screenshots

image

image


nft-studio/src/components/CreateCollectionDialog.tsx
```diff
@@ -29,6 +29,8 @@ export const CreateCollectionDialog: React.FC<{ open: boolean; onClose: () => vo
   const cent = useCentrifuge()
   const { data: balance } = useBalance()
   const [redirect, setRedirect] = useState<string>('')
+  const [confirmOpen, setConfirmOpen] = useState(false)
+  const [termsAccepted, setTermsAccepted] = useState(false)
```
alessioscalici, line 33: Maybe I'm missing it in the code, but shouldn't we store this in the localstorage, to avoid the user having to accept T&C every time he/she creates a collection?

onnovisser: Agree that that would be nice, but it isn't in the spec

jpangelle (approved):
:shipit:

PR #744, opened 2022-03-21
NFT Playground: Save collection logo when switching dialogs

onnovisser:
### Description

When in the Create Collection dialog and going to the Terms dialog and back, the collection logo wasn't being saved. Changes the Fabric FileUpload component so that it's state can optionally be controlled via props.

### Approvals

- [x] Dev

### Impact

NFT Playground, Fabric


jpangelle (approved):
nice!

PR #746, opened 2022-03-24
Make mailer service injectible and trigger agreement email

avinashega:
### Description

This pull request triggers subscription agreement email in both agreement callback and docusign connect webhook

Closes #<issue-number>

### Approvals

- [ ] Dev

### DIFF
--- a/onboarding/api/src/services/mailer.service.ts
+++ b/onboarding/api/src/services/mailer.service.ts
@@ -2,68 +2,80 @@ import { Injectable } from '@nestjs/common'
 import config from '../config'
 import { User } from '../repos/user.repo'
 
-const fetch = require('@vercel/fetch-retry')(require('node-fetch'))
+const client = require('@sendgrid/mail')
+client.setApiKey(config.sendgrid.apiKey)
 
 @Injectable()
 export class MailerService {
   async sendWhitelistedEmail(user: User, pool: any, data: any) {
     const issuerName = pool.profile?.issuer?.name.replace(/\s+/g, '-').toLowerCase()
-    const response = await fetch(config.sendgrid.apiUrl, {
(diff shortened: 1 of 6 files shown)


onnovisser (approved):

onboarding/api/src/services/mailer.service.ts
```diff
@@ -0,0 +1,82 @@
...
+      },
+      template_id: config.sendgrid.whitelistEmailTemplate,
+    }
+
+    await client
+      .send(message)
+      .then(() => console.log('Subscription email sent successfully'))
+      .catch((error) => {
+        console.error(error)
+      })
```
jpangelle, line 45: curious on why the mixture of `async/await` and `then` rather than:
```ts
try {
  await client.send(message);
  console.log('Subscription email sent successfully');
} catch (error) {
  console.error(error);
}

```

onboarding/api/src/services/mailer.service.ts
```diff
@@ -0,0 +1,82 @@
...
+      },
+      template_id: config.sendgrid.subscriptionAgreementTemplate,
+    }
+
+    await client
+      .send(message)
+      .then(() => console.log('Mail sent successfully'))
+      .catch((error) => {
+        console.error(JSON.stringify(error))
+      })
```
jpangelle, line 78: same here

jpangelle (commented):
lgtm, was just wondering on the mix of `async/await` and `then`

onboarding/api/src/services/mailer.service.ts
avinashega: Coming from a different school of thought. But I agree, try/catch is more readable. #750 has the updated code.

PR #755, opened 2022-03-29
Add versions to tinlake params so bot will correctly call NAV feed functions

AStox:
### Description

This pull request adds the contract versions metadata to the bots tinlake.js instance so that it can call the correct NAVfeed functions. This was bug was causing any contract with a FEED version of 2 to error and not print to any slack channel.

This PR also edits a query to filter loan assets with a debt value of 0.

Closes #<issue-number>

### Approvals

- [x] Dev


jpangelle (approved):
🤖

PR #803, opened 2022-05-04
Centrifuge App: Issuer Investors Tab

onnovisser:
### Description

Adds the tab to the issuers page to add/remove investors to tranches

Also removes epoch time from pool creation

### Approvals

- [ ] Dev

### Screenshots

image
image
image

### Impact

Centrifuge App


sophialittlejohn (approved):
I can confirm pool creation works again the the new investor tab looks good and works well 🏅

Issue #704, opened 2022-03-10
List of Assets - Public Pools -> Assets page MVPv1


EstherCentrifuge:
[Specs ](https://centrifuge.hackmd.io/MSB1Ll8IRfeXPmukXMeBtw#List-of-Assets)
### List of Assets
Lists all assets within this pool.
The list can be sorted by any column within the table
Each asset shows 
- Dimond shaped icon with asset order number
- Asset Name
    [- Set during asset creation](https://centrifuge.hackmd.io/2NJLoV7yRXSOC2YrrImXCg)
- Maturity
    - maturity date - origination date = Maturity
- Maturity date
    - [Set during asset creation](https://centrifuge.hackmd.io/2NJLoV7yRXSOC2YrrImXCg)
- Amount
     - Collateral value [Set during asset creation](https://centrifuge.hackmd.io/2NJLoV7yRXSOC2YrrImXCg)
- Status of the asset
[See Specs here](https://centrifuge.hackmd.io/sUnY0kNoQamfj9I_IFquoQ)

sophialittlejohn:
@EstherCentrifuge should this list only display assets of type `CreditLineWithMaturity`?

sophialittlejohn:
resolved: it should display all asset with all status', thanks @EstherCentrifuge

EstherCentrifuge:
closed as there are updaets in a seperate ticket but the feature it self works

PR #818, opened 2022-05-11
Pool detail assets tab

sophialittlejohn:
### Description

Adds asset tab in public pool detail, for example `pools/3075481758/assets`

Linked #704 

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product

### Screenshots

Screenshot 2022-05-11 at 16 34 18

### Impact

fabric (just the theme)
centrifuge-app (nft-studio)

### DIFF
--- a/fabric/src/components/Thumbnail/index.tsx
+++ b/fabric/src/components/Thumbnail/index.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import styled, { css } from 'styled-components'
+import styled, { css, useTheme } from 'styled-components'
 import { Stack } from '../Stack'
 import { Text } from '../Text'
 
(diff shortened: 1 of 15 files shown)


fabric/src/components/Thumbnail/index.tsx
```diff
@@ -46,7 +46,7 @@ const StyledThumbnail = styled(Text)<Partial<ThumbnailProps>>`
             bottom: 0,
             right: 0,
             margin: 'auto',
-            zIndex: 0,
+            zIndex: 'base',
```
onnovisser, line 49: I'm not sure it makes sense to have the -1, 0, and 1 values as part of the theme. These are very local. Editing via the theme can only break the layout

sophialittlejohn: true... I guess those three are just for consistency. Would it make sense to have some defined in the theme and some not?

onnovisser: I think that's fine

onnovisser (approved):

onnovisser:
🙌

PR #821, opened 2022-05-13
Centrifuge App: Update chart styles

sophialittlejohn:
### Description

This pull request... 
- fixes charts according to updated [design](https://www.figma.com/file/d1bEBkwllUQe27nQM6VlJu/Centrifuge-app---Version-1?node-id=4235%3A98485) 
- removes portfolio tab from issuer pools
- removes maturity column from asset when asset is unpriced and not borrow against (if I don't do this origination date is set to 1970 so we get obscure maturities)

Please note:
I wasn't able to use the font-variants that were requested (slashed zeros). Our font Inter doesn't support it.
You can test it on the[ google font page for Inter](https://fonts.google.com/specimen/Inter?preview.text=012344&preview.text_type=custom) by adding the following CSS to a test element
```css
font-variant-numeric: slashed-zero;
```

Closes #<issue-number>

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product

### DIFF
--- a/nft-studio/src/components/LoanList.tsx
+++ b/nft-studio/src/components/LoanList.tsx
@@ -27,7 +27,9 @@ export const LoanList: React.FC<Props> = ({ loans, onLoanClicked }) => {
       header: 'Maturity',
       cell: (l: Loan) => (
         <Text variant="body2">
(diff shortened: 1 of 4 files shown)


onnovisser:
image

PR #826, opened 2022-05-18
Centrifuge App: Design feedback

onnovisser:
### Description

Implement a bunch of design feedback. 

- Keep submit buttons enabled, even when there are errors. adds hook to focus invalid input upon submission
- Fix asset table overflow
- Add back buttons to page headers
- Change nft pages paths to make the Nfts menu item highlighted when on those pages
- Text tweaks

### Approvals

- [ ] Dev

### Impact

Centrifuge App


sophialittlejohn (approved):
🥳

PR #836, opened 2022-05-21
Minor Centrifuge App UI fixes

offerijns:
### Description
A few bug fixes:
- Fix the current protection calculations.
- Fix the average maturity.
- Fix decimals of wallet balance.
- Use actual native token symbol rather than hardcoded AIR.
- Fix the currency of the TVL on the token list.

And a few improvements:
- Simplify navigation (remove unused pages).
- Use new icons in navigation.
- Improve portfolio page according to latest design.
- Add a network env var and make some things dependent on whether the target is Altair or Centrifuge.
- Add a few sample asset classes for Centrifuge.
- Token symbols should be 4-12 chars and share the first 3 chars.
- Add ability to manually edit pool config.

### Approvals

- [ ] Dev
- [ ] Dev


centrifuge-js/src/modules/pools.ts
```diff
@@ -188,11 +188,10 @@ export type Tranche = {
   index: number
   id: string
   seniority: number
-  debt: Balance
```
offerijns, line 191: I think we shouldn't be exposing `debt`, `reserve`, or `ratio` to avoid potential issues; there's really no reason a client would need these, they're just properties used for bookkeeping on chain

nft-studio/src/utils/useCurrencies.ts
```diff
@@ -0,0 +1,22 @@
+import { useCentrifugeQuery } from './useCentrifugeQuery'
+
+export function useCurrencies(address?: string) {
+  const [result] = useCentrifugeQuery(['balances', address], (cent) => cent.pools.getBalances([address!]), {
+    enabled: !!address,
+  })
+
+  return [
```
offerijns, line 8: We'll add an RPC endpoint in the future to retrieve these from chain.

centrifuge-js/src/modules/pools.ts
```diff
@@ -786,15 +797,32 @@ export function getPoolsModule(inst: CentrifugeBase) {
                 minRiskBuffer = new Perquintill(hexToBN(tranche.trancheType.nonResidual.minRiskBuffer))
                 interestRatePerSec = new Rate(hexToBN(tranche.trancheType.nonResidual.interestRatePerSec))
               }
+
+              const subordinateTranchesValue = pool.tranches.tranches
```
sophialittlejohn, line 801: what's the significance of a `subordinateTrancheValue`, i.e what makes a tranche a subordinate? Just trying to understand how it realted to the current risk buffer

nft-studio/src/components/TokenList.tsx
```diff
@@ -38,18 +38,18 @@ const columns: Column[] = [
...
     cell: (token: TokenTableData) => <Yield token={token} />,
-    flex: '4',
+    flex: '3',
     sortKey: 'yield',
   },
   {
     header: () => <SortableTableHeader label="Protection" />,
     cell: (token: TokenTableData) => <Protection token={token} />,
-    flex: '4',
+    flex: '3',
```
sophialittlejohn, line 52: FYI on my 14'' screen this has caused the `Protection` column to wrap

sophialittlejohn (approved):
The app is starting to feel really nice to use 😍  Nice job 🙌  I left just a few thoughts/suggestions

Also, I like the reduced navigation, it saves a lot of duplicication (which probably also means there are a few more files to delete)

For the future I think we will need to come up with a scalable system to switch between altair and centrifuge, maybe some sort of config

onnovisser (approved):
Nice!

centrifuge-js/src/modules/pools.ts
offerijns: Subordinate meaning lower in seniority. The protection / risk buffer is the value of all the tranches with less seniority, i.e. those protecting the current tranche from losses.

EstherCentrifuge:
Thanks for the bug fixes and some(most) of the improvements.

two things only.
1. please put back the pool list We can make an investment only menu but then we need that switch to the pool list as well and it's a different design happy to do that but it's not a "bug" for Altair lets have tokens and pool seperate for now. we need a pool list some how. 

2. Token symbols should be 4-12 chars and share the first 3 chars.
I saw the deiscussion i think it makes sense but we need to see the impact around the application as we use the "symbols" in the token icon and lists so if they get long we need to cut them off. I suggest if we go this way we cut off the start not the end (I am sure this was in the specs anyway) to remove the fixed pool identifying bit so make sure they don't all look the same.

Issue #839, opened 2022-05-23
Bug - Risk groups not shown correct data in pie chart


EstherCentrifuge:
**example 1:** Esther's Pool 
Risk group 5 = 0 amount 0% share
Risk group 10 = 0 amount 0% share
Total = 87
https://dev-nft-studio.netlify.app/issuer/3189422571

![image](https://user-images.githubusercontent.com/98879480/169853367-bc440ed8-ebd0-4e2a-845a-7475ad1431db.png)
But there are outstanding assets in Risk group 10:
https://dev-nft-studio.netlify.app/pools/3189422571/assets/9
https://dev-nft-studio.netlify.app/pools/3189422571/assets/19 

**Example 2:**  Cool Pool
https://dev-nft-studio.netlify.app/pools/3075481758/assets/5
Asset 5 = risk Group C -> 141 outstanding

https://dev-nft-studio.netlify.app/pools/3075481758
Piechart states Risk group C = 0 amount 0% share
![image](https://user-images.githubusercontent.com/98879480/169853595-2830e11c-c7bf-4218-a9bc-0bfa95c7e040.png)

EstherCentrifuge:
we spoke about this, due to the "other" group it's not showing correct but without it it works so I'll close it

PR #842, opened 2022-05-23
Centriuge App: Fix RiskGroup Pie Chart

sophialittlejohn:
### Description

This pull request...
PieChart now supports all loan types 

There are 2 examples (so far) where the amount/shares in the table/chart are incorrect. I double-checked the math and code and the only problem I came up with is that the NAV may be incorrectly calculated.
The `amount` column is the sum of all outstanding debts per risk group (after sorting loans into risk groups)
The `share` column is the `amount / totalNAV`
The `Total amount` reflects the current nav on-chain

**Example 1**
Based on the share percentage of risk group **5** and **10** either the NAV or the outstanding debt must be incorrect on-chain. (**Other** is just a catch-all to make up for any assets that cannot definitely be assigned to a risk group)
Screenshot 2022-05-23 at 19 47 36

**Example 2**
The risk group `Medium risk` has outstanding debt of ~100$ but the NAV on chain is 0
Screenshot 2022-05-23 at 19 54 10

Links #839

### Approvals

- [ ] Dev


sophialittlejohn:
Apparently, the NAV is off because the maturity date is too far in the future. This means that the present value is negative. Will merge so that the rest of the pies display nicely while product works on a solution.

PR #847, opened 2022-05-24
Centrifuge App: Tranche metadata update

onnovisser:
### Description

Saves tranche metadata by trancheId and updates the UI to expect it in that way (breaks tranche metadata for all pools until they are resaved). Also persists proxy address to localstorage.

### Approvals

- [ ] Dev

### Impact

Centrifuge App

### DIFF
--- a/nft-studio/src/components/Web3Provider.tsx
+++ b/nft-studio/src/components/Web3Provider.tsx
@@ -60,10 +60,13 @@ export const Web3Provider: React.FC = ({ children }) => {
 
     setAccounts(kusamaAccounts)
     const persistedAddress = localStorage.getItem('web3PersistedAddress')
-    const address =
-      (persistedAddress && kusamaAccounts.find((acc) => acc.address === persistedAddress)?.address) ||
-      kusamaAccounts[0]?.address
+    const persistedProxy = localStorage.getItem('web3PersistedProxy')
+    const matchingAccount = persistedAddress && kusamaAccounts.find((acc) => acc.address === persistedAddress)?.address
+    const address = matchingAccount || kusamaAccounts[0]?.address
     setSelectedAccountAddress(address)
+    if (matchingAccount && persistedProxy) {
+      setProxyAddress(persistedProxy)
+    }
     localStorage.setItem('web3PersistedAddress', address ?? '')
   }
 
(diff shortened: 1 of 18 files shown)


nft-studio/src/pages/IssuerCreatePool/pinPoolMetadata.ts
```diff
@@ -1,10 +1,31 @@
...
+  }
+  return `0x${out.join('')}`
+}
+
+function computeTrancheId(trancheIndex: number, poolId: string) {
+  const a = new BN(trancheIndex).toArray('le', 8)
+  const b = new BN(poolId).toArray('le', 8)
+  const data = Uint8Array.from(a.concat(b))
+
+  return toHex(hash(data, 16))
```
sophialittlejohn, line 26: nice job figuring this out 🎉

nft-studio/src/components/Web3Provider.tsx
```diff
@@ -115,6 +120,11 @@ export const Web3Provider: React.FC = ({ children }) => {
     setProxyAddress(null)
   }, [])
 
+  const selectProxy = React.useCallback((address: string | null) => {
+    setProxyAddress(address)
+    localStorage.setItem('web3PersistedProxy', address ?? '')
```
sophialittlejohn, line 125: Just this seems a bit weird to me. I see we're doing it in other places as well. But why are we potentially setting `''` into localStorage? I would think that we shouldn't set the item if `address === null`

sophialittlejohn (approved):
nice job 💪  Let's make sure we update the all of metadata quickly once this is merged!

nft-studio/src/components/Web3Provider.tsx
onnovisser: if we're calling `selectProxy` with `null`, we'd need to clear any persisted proxy. the alternative would be
```
if (address) {
    localStorage.setItem('web3PersistedProxy', address)
} else {
    localStorage.removeItem('web3PersistedProxy')
}
```
the current is just a bit more concise and the component treats the empty strings it finds in localStorage as undefined. If you think the more verbose alternative is better/more readable, I'm happy too change it.

sophialittlejohn: fair enough, for me it's fine as it 😄

PR #853, opened 2022-05-30
Simluation tests update

sophialittlejohn:
### Description

This pull request... fixes simulations tests and updates them with better (more realistic) data. It also improves the remaining epoch time calculation and fixes some minor UI problems.

### Approvals

- [ ] Dev


nft-studio/src/pages/PoolDetail/LiquidityTab/index.tsx
```diff
@@ -67,7 +68,7 @@ export const PoolDetailLiquidity: React.FC = () => {
         titleAddition={`Ongoing`}
         headerRight={
           <Shelf gap="1">
-            <Tooltips type="epochTimeRemaining" label={`${getEpochHoursRemaining(pool!)} hrs remaining`} />
+            <Tooltips type="epochTimeRemaining" label={`${hours} hrs and ${minutes} min remaining`} />
```
onnovisser, line 71: looks like `minutes` currently isn't the remainder, but the whole time remaining in minutes

nft-studio/src/components/InvestRedeem.tsx
```diff
@@ -568,7 +569,7 @@ const PendingOrder: React.FC<{
           </Shelf>
           <Text variant="body3">
             Locked {type === 'invest' ? 'investments' : 'redemptions'} are executed at the end of the epoch (
-            {getEpochHoursRemaining(pool!)} hrs remaining).{' '}
+            {hoursRemaining} hrs remaining).{' '}
```
onnovisser, line 572: I feel this would be better rounded up to the nearest hour, otherwise it'll say `0 hours remaining` during the whole last hour

simulation-tests/src/main.ts
```diff
@@ -1,155 +1,198 @@
-import Centrifuge from '@centrifuge/centrifuge-js'
+import Centrifuge, { Balance, LoanInfoInput, Perquintill, Pool, Rate } from '@centrifuge/centrifuge-js'
```
onnovisser, line 1: Looking good!

nft-studio/src/components/InvestRedeem.tsx
sophialittlejohn: hmm yeah, that's not good. It seems it would make the most sense to show minutes here as well rather than rounding hours in one place and being specific in the other 🤔

onnovisser (approved):

Issue #852, opened 2022-05-30
UI fixes to table component: Row height


armandomistral:
#### Is
The row-height of table rows (both table header rows and table body rows) is not consistent, and too low.

![image.png](https://images.zenhubusercontent.com/6090f1ca0d0c23544195c7ec/6701418e-1c16-47fa-a941-d145d8e24bfd)

![image.png](https://images.zenhubusercontent.com/6090f1ca0d0c23544195c7ec/b616276e-5363-474c-872b-3296b1a43f58)

#### Target
* Apply consistent height of 48px to both header and body rows
* Make table rows consistent in height across instances, i.e.:
  * Token menu
  * Pool menu
  * Investment tokens (in pool detail)
  * Tokens (in Portfolio)
* [Component spec](https://www.figma.com/file/TPFzwr1VVtg7MZlRVwMKvS/Fabric-design-system?node-id=988%3A3649)
* [Token menu spec](https://www.figma.com/file/LCbVWH8fXPkXMbLpC3VZtx/Centrifuge-app---Altair?node-id=3692%3A83628)

![image.png](https://images.zenhubusercontent.com/6090f1ca0d0c23544195c7ec/6624f4d0-2de2-4596-b8f6-610ab01c3629)

![image.png](https://images.zenhubusercontent.com/6090f1ca0d0c23544195c7ec/0e1ca0b9-dc11-4ba2-b38c-589232ab6981)

sophialittlejohn:
@armandomistral how should we handle words that are too long? Currently, they wrap lines. We could add an ellipsis (...) or a scroll overflow. Please let me know 🙏

sophialittlejohn:
yeah, let's go with an ellipsis 🙌  and yep, you can always zoom out to see the full content

armandomistral:
Thanks for the fixes, @sophialittlejohn! Tested on https://dev.app.altair.cntrfg.com/:
- Table row height fixed to 48px: ✅
- Preventing cell text from wrapping: ✅
- Ellipsing content:
   * Token name column of token table: ✅
   * Asset class column of token table: ⚠️ Not set
… Do you have an idea on how we might standardize the ellipsing of content so that it's applied to all columns? So we don't need to it manually on a case-by-case basis...

Issue #851, opened 2022-05-26
fix altair colour scheme


EstherCentrifuge:
currently the active items (menu, tabs, titles etc.)  are blue (as it should be in the white design)
On Altair it should be Yellow. see [Figma](https://www.figma.com/file/LCbVWH8fXPkXMbLpC3VZtx/Centrifuge-app---Altair?node-id=4440%3A100305)

Example AS IS:
![image.png](https://images.zenhubusercontent.com/61fa54b8b520bc58fd924e32/168e00f2-2f48-47ae-8019-c256a31c6cca)

Example Target:
![image.png](https://images.zenhubusercontent.com/61fa54b8b520bc58fd924e32/739ed123-8c5f-4a49-87be-cbaa00ef5e37)

armandomistral:
Issues spotted while testing the dark theme on https://dev.app.cntrfg.com/. Can't take screenshots anymore since all data got cleared, but here's the list:
- **Table**: Background color is dark gray but should be backgroundPage (black)
- **Investment component**: Background color is dark gray but should be backgroundPage
- **Tooltip indicator**: Color is white but should be the label color (textSecondary)

PR #863, opened 2022-06-02
Centrifuge App: Altair leftover clean up

sophialittlejohn:
### Description

This pull request...

- adds more altair color improvements
- lazy loads chart components for better chunking

#851 
#852 (partial - currently overflow in tables is an ellispis)

### Approvals

- [ ] Dev


centrifuge-app/src/components/Web3Provider.tsx
```diff
@@ -96,7 +96,7 @@ export const Web3Provider: React.FC = ({ children }) => {
       if (!wallet?.installed) throw new Error('Wallet not available')
       setSelectedWallet(wallet)
 
-      // const injected = await (web3EnablePromise || web3Enable('NFT Studio'))
+      await (web3EnablePromise || web3Enable('Centrifuge App'))
```
sophialittlejohn, line 99: @onnovisser I'm not sure why this was commented out, but I needed it to sign transactions

onnovisser: It was replaced by `wallet.enable(config.name)`, so it should actually have been removed. What happens when you try doing an transaction without it?

sophialittlejohn: This was the error:
Screenshot 2022-06-01 at 18 59 33

fabric/src/components/Tooltip/index.tsx
```diff
@@ -19,11 +19,23 @@ const StyledTrigger = styled(Text)`
...
   position: relative;
-  text-decoration: underline dotted;
   cursor: pointer;
   text-align: left;
   font-weight: 400;
   font-family: ${({ theme }) => theme.fonts.standard};
+  position: relative;
+  width: fit-content;
+
+  ::after {
```
onnovisser, line 29: can use `text-decoration-color` instead. positioning the line absolutely will look a bit weird if the text wraps.

sophialittlejohn: the problem wasn't actually the color but the space in between the text and the underline, that's why I had to do it with the absolute positioning. Maybe we can prevent wrapping instead? Use the ellipsis here as well 🤔

centrifuge-app/src/components/Web3Provider.tsx
sophialittlejohn: Seems to be working after all 🤯  not sure why, but line has been removed again

Issue #874, opened 2022-06-15
Bug: Remove equal sign in Pricing section of asset detail


armandomistral:
## Is
In the **Pricing** section of the asset detail, there is a equal sign ("=") between `Expected loss` and `Discount rate`, which is not correct.

![image.png](https://images.zenhubusercontent.com/6090f1ca0d0c23544195c7ec/0fe2792d-8d87-4d7b-83a3-8053f879b87a)

## Target
Remove the equal sign character between `Expected loss` and `Discount rate`.

armandomistral:
Thanks, @sophialittlejohn! Checked on https://dev.app.altair.cntrfg.com/, it looks good
<img width="836" alt="image" src="https://user-images.githubusercontent.com/23652850/175017271-4b4a5766-ba6e-4e5b-8095-215114f14d0f.png">

Issue #846, opened 2022-05-24
Asset repayment


armandomistral:
## Is
> Currently, the subtext of the repayment input shows the pool currency balance. Instead, it should the minimum of the pool currency balance and the outstanding amount (i.e. if outstanding < balance, then the subtext should just show the outstanding, as this is the max that can be repaid)
[Slack](https://centrifugehq.slack.com/archives/C033T5YUX0W/p1653165813582759)

![image](https://user-images.githubusercontent.com/23652850/170033650-c1780cfc-fcfc-4be4-9977-87c739719fa8.png)

## Target
* Available amount in currency input field: minimum value of `wallet balance` and `outstanding amount`
* Max: enters available amount
* Repay all: disabled if available < outstanding
* Show info message if `wallet balance` is shown

Example: `Your wallet balance (589 kUSD) is smaller than the outstanding`

[Figma](https://www.figma.com/file/d1bEBkwllUQe27nQM6VlJu/Centrifuge-app---Version-1-(Master)?node-id=7063%3A120062)

![image](https://user-images.githubusercontent.com/23652850/170266029-79d03eaa-425e-4172-ad5b-995674373a26.png)

#### Related ticket
https://github.com/centrifuge/apps/issues/844

EstherCentrifuge:
It should show the issuers wallet balance.
Ticket will be updated by @armandomistral with the new design :) 
Thank!

armandomistral:
Done, @EstherCentrifuge

EstherCentrifuge:
@onnovisser 
We should add thousand separators to the currency inputs. -> comma thousand, point decimal

Quickly checked, we used to use this library for it in Tinlake UI: https://github.com/s-yadav/react-number-format#readme. But there’s probably simpler ways to implement it, e.g.: https://stackoverflow.com/questions/63091317/thousand-separator-input-with-react-hooks

sophialittlejohn:
An issue that came up during testing: it appears that outstanding value doesn't include accrued interest, I will double-check

Issue #844, opened 2022-05-24
Asset financing


EstherCentrifuge:
## Is
* The input amount does not show a balance
* The max button fills the available financing amount

![image](https://user-images.githubusercontent.com/23652850/170032882-e18bf220-ecf7-4f08-a711-0996573f68fd.png)

## Target
> And for the borrow component, the subtext should just show the minimum of the available financing and the available reserve.
> [Slack](https://centrifugehq.slack.com/archives/C033T5YUX0W/p1653165833003329)

* Available amount: minimum value of `available financing`, `available reserve`
    * Available pool reserve - my (the issuer) pending repayments (in epoch)= available reserve for the issuer
        * After Epoch is closed/executed my pending repayments = 0 (as nothing is pending) 
* Max: enters available amount
* Show the balance amount depending on available financing and available reserve:
    * Show the minimum of the `available financing` and the `available reserve`

Example: `589 kUSD available`

[Figma: part - finance asset - exception](https://www.figma.com/file/d1bEBkwllUQe27nQM6VlJu/Centrifuge-app---Version-1-(Master)?node-id=7075%3A115344)

![image](https://user-images.githubusercontent.com/98879480/170258412-de77cdfe-8a87-49dd-808d-22d5fa712854.png)

#### The max amount 
The max amount to finance is “flexible” please update to exact with 2 decimals
Example in total I can borrow 140 against my asset, i borrow 100, now 5 min later i can not borrow 40 becuase of the maturity and interest, this is logic and works well.
However the “available financing” states 40 but then when i fill 40 it says it exceeds the available financing. hwne you press the “max” option it gives the correct amount. Digets behind comma for “max amount” always roundned down (so 97,8767654 = 97,87)

![image.png](https://images.zenhubusercontent.com/61fa54b8b520bc58fd924e32/0a5a63e5-3ac5-41af-8e6e-c6c5adbc1dc1)

EstherCentrifuge:
@onnovisser  see my comment above, , I know i wrote it somewhere :)

sophialittlejohn:
@EstherCentrifuge @armandomistral regarding the available amount
>The max amount to finance is “flexible” please update to exact with 2 decimals
Example in total I can borrow 140 against my asset, i borrow 100, now 5 min later i can not borrow 40 becuase of the maturity and interest, this is logic and works well.
[...]

PR #875, opened 2022-06-15
Centrifuge App: Fix finance and repayment forms

sophialittlejohn:
### Description

This pull request...
- updates all currency inputs with thousand separator
    - invest form (with max button)
    - redeem form (with max button)
    - finance form (with max button)
    - repay form (with max button)
    - create pool --> initial max reserve field
    - price asset --> collateral value field
    - sell asset --> price field
- rounds max button input down to 2 decimal places
- adds decimals to invest/redeem/finance/repayment forms to avoid losing precision on max button press
- fixes some broken (back)links
- removes the `=` in the pricing form (#874)
- implements change requests in #846 and #844
- allows financing after the asset has been fully financed and repaid (#886) 

#846 
#844
#874
#886 

### Approvals

- [ ] Dev

### DIFF
--- a/centrifuge-app/src/pages/Loan/FinanceForm.tsx
+++ b/centrifuge-app/src/pages/Loan/FinanceForm.tsx
@@ -64,7 +64,7 @@ export const FinanceForm: React.VFC<{ loan: LoanType }> = ({ loan }) => {
   )
   const poolReserve = pool?.reserve.available.toDecimal() ?? Dec(0)
   const maxBorrow = poolReserve.lessThan(availableFinancing) ? poolReserve : availableFinancing
-  const maxRepay = Math.min(balance.toNumber(), loan.outstandingDebt.toDecimal().toNumber())
+  const maxRepay = balance.lessThan(loan.outstandingDebt.toDecimal()) ? balance : loan.outstandingDebt.toDecimal()
   const canRepayAll = debtWithMargin.lte(balance)
 
   const financeForm = useFormik<FinanceValues>({
@@ -103,11 +103,12 @@ export const FinanceForm: React.VFC<{ loan: LoanType }> = ({ loan }) => {
         <Stack>
           <Shelf justifyContent="space-between">
             <Text variant="heading3">Available financing</Text>
+            {/* availableFinancing needs to be rounded down, b/c onSetMax displays the rounded down value as well */}
             <Text variant="heading3">{formatBalance(roundDown(availableFinancing), pool?.currency, 2)}</Text>
           </Shelf>
           <Shelf justifyContent="space-between">
             <Text variant="label1">Total financed</Text>
-            <Text variant="label1">{formatBalance(roundDown(loan.totalBorrowed.toDecimal()), pool?.currency, 2)}</Text>
+            <Text variant="label1">{formatBalance(loan.totalBorrowed.toDecimal(), pool?.currency, 2)}</Text>
           </Shelf>
         </Stack>
         {loan.status === 'Active' && loan.totalBorrowed.toDecimal().lt(initialCeiling) && (
@@ -133,7 +134,7 @@ export const FinanceForm: React.VFC<{ loan: LoanType }> = ({ loan }) => {
                     disabled={isFinanceLoading}
                     currency={getCurrencySymbol(pool?.currency)}
                     handleChange={(value: number) => form.setFieldValue('amount', value)}
-                    onSetMax={() => form.setFieldValue('amount', roundDown(maxBorrow))}
+                    onSetMax={() => form.setFieldValue('amount', maxBorrow)}
                   />
                 )}
               </Field>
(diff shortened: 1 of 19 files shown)


centrifuge-app/src/utils/formatting.ts
```diff
@@ -48,3 +48,7 @@ export function formatPercentage(amount: Perquintill | Decimal | number, include
   })
   return includeSymbol ? `${formattedAmount}%` : formattedAmount
 }
+
+export function roundDown(float: Decimal | number) {
+  return Math.floor((float instanceof Decimal ? float.toNumber() : float) * 100) / 100
+}
```
offerijns, line 54: This should probably accept a precision. Now it's implicitly rounding to a precision of 2 decimals. Something like:

```suggestion
export function roundDown(float: Decimal | number, precision: number = 2) {
  return Math.floor((float instanceof Decimal ? float.toNumber() : float) * 10**precision) / 10**precision
}
```

fabric/src/components/CurrencyInput/index.tsx
```diff
@@ -9,9 +9,12 @@ export type CurrencyInputProps = React.InputHTMLAttributes<HTMLInputElement> &
   Omit<InputBoxProps, 'inputElement' | 'rightElement'> & {
     currency?: string
     onSetMax?: () => void
+    variant?: 'small' | 'large'
+    handleChange?: (value: number) => void
```
offerijns, line 13: I think `onChange` is more standard, and in line with `onSetMax`?

```suggestion
    onChange?: (value: number) => void
```

fabric/src/components/CurrencyInput/index.tsx
```diff
@@ -68,8 +91,26 @@ export const CurrencyInput: React.FC<CurrencyInputProps> = ({
...
+    const inputFormatted = formatThousandSeparator(value)
+    const inputAsNumber = parseFloat(value.replaceAll(',', ''))
+    if (inputProps?.handleChange) {
+      inputProps?.handleChange(inputAsNumber)
+      setValue(inputFormatted)
+    }
+  }
+
+  React.useEffect(() => {
+    const inputFormatted = formatThousandSeparator(Math.floor((inputProps.value as number) * 100) / 100)
```
offerijns, line 110: Similar to `roundDown`, the decimals shown should be configurable, using a `decimals` argument on `CurrencyInput`.

offerijns (changes requested):
Looks great! Tested the thousand separators in the preview, works perfectly 🔥 Just left a few minor comments.

fabric/src/components/CurrencyInput/index.tsx
sophialittlejohn: good call, we would definitely need it sooner or later 👍  I defaulted the precisions to `2` everywhere

centrifuge-app/src/pages/Loan/FinanceForm.tsx
```diff
@@ -67,6 +69,16 @@ export const FinanceForm: React.VFC<{ loan: LoanType }> = ({ loan }) => {
   const maxRepay = balance.lessThan(loan.outstandingDebt.toDecimal()) ? balance : loan.outstandingDebt.toDecimal()
   const canRepayAll = debtWithMargin.lte(balance)
 
+  const allowedToBorrow: Record<LoanInfo['type'], boolean> = {
```
offerijns, line 72: I'm not sure I follow this, why is this always `false` for bullet loans and credit lines?

sophialittlejohn: I added validation for BulletLoans and CreditLines. Can you please double check the logic? It seems a little too simple the way it is now...

centrifuge-app/src/pages/Loan/FinanceForm.tsx
```diff
@@ -65,8 +66,21 @@ export const FinanceForm: React.VFC<{ loan: LoanType }> = ({ loan }) => {
...
+    CreditLineWithMaturity:
+      'maturityDate' in loan.loanInfo &&
+      Dec(daysBetween(loan.originationDate, loan.loanInfo.maturityDate)).gt(0) &&
+      loan.outstandingDebt.toDecimal().lt(availableFinancing),
+    BulletLoan:
+      'maturityDate' in loan.loanInfo &&
+      Dec(daysBetween(loan.originationDate, loan.loanInfo.maturityDate)).gt(0) &&
+      loan.totalRepaid.toDecimal().lt(initialCeiling),
```
offerijns, line 80: I think this should be:
```suggestion
      loan.totalBorrowed.toDecimal().lt(initialCeiling),
```

offerijns (changes requested):

centrifuge-app/src/pages/Loan/FinanceForm.tsx
sophialittlejohn: good point, just noticed that with `totalRepaid` the form would still be open even if the `maxBorrowed` is reached

offerijns (approved):
LGTM 👍

PR #877, opened 2022-06-16
Centrifuge App: Get token prices via RPC

onnovisser:
### Description

Get token prices via RPC so they're more up to date. We currently get them from the latest epoch, so it's usually out of date. It needs to do a rpc call per pool. It does this in parallel with the rest of the pool data, but allows the pool to emit it's first value without having to wait for all the requests to finish. A side-effect of this is that `tokenPrice` can be `null` in the returned data and you can visually see the TVL increase as data loads on the Pools/Tokens page.

### Approvals

- [ ] Dev

### Impact

CentrifugeJS, Centrifuge App


sophialittlejohn (approved):
💯

PR #882, opened 2022-06-17
Add solver setup

offerijns:
### Description

This pull request... adds solver to the UI. Currently, the solver can only provide solutions if the available reserve is too low to execute all order.

To test:
1. start with a pool at `Pool Value === 0`
2. set max reserve to 10'000
3. invest 6000 in junior tranche
4. invest 6000 in senior tranche
5. close epoch --> will go into submission period
6. submit solution and wait for challenge time to pass (will take 10 blocks)
7. execute the epoch after challenge time passes. If successful, the pool value should be the same as the max reserve.

- #883 

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product

### Screenshots

close-epoch
submission-period
challenge-period
execution-period


EstherCentrifuge:
@sophialittlejohn what's this ticket? and can we make the status up to date

centrifuge-app/src/components/EpochList.tsx
```diff
@@ -35,19 +36,21 @@ const columns: Column[] = [
...
   const { data: metadata } = usePoolMetadata(pool)
 
-  const investments: TableDataRow[] = pool.tranches.map((token) => {
+  const investments: TableDataRow[] = pool.tranches.map((token, index) => {
     const trancheMeta = metadata?.tranches?.[token.id]
+    const price = token.tokenPrice?.toDecimal() || Dec(0)
     return {
       order: `${trancheMeta?.symbol} investments`,
-      locked: token.outstandingInvestOrders?.toDecimal() || new Decimal(0),
+      locked: token.outstandingInvestOrders?.toDecimal().mul(price) || new Decimal(0),
```
offerijns, line 44: I think it would be good to add a `toTokenBalance(price)` method to `CurrencyBalance`, so this can be changed to
```suggestion
      locked: token.outstandingInvestOrders?.toTokenBalance(price).toDecimal() || new Decimal(0),
```

centrifuge-app/src/pages/Pool/Liquidity/index.tsx
```diff
@@ -55,18 +57,91 @@ export const PoolDetailLiquidity: React.FC = () => {
...
+        invest: new BN(10).pow(new BN(solutionTranches.length - index)),
+        redeem: redeemStartWeight.mul(new BN(10).pow(new BN(index).addn(1))),
+      }
+    })
+
+    const solution = await calculateOptimalSolution(poolState, orders, weights)
+    if (solution.isFeasible) {
+      submitSolutionTx([pool.id, solution.tranches])
+    }
+  }
```
offerijns, line 141: I think all of this could be moved into `submitSolution`, so it just has a `poolId` argument and everything else is calculated in there.

centrifuge-app/src/pages/Pool/Liquidity/index.tsx
```diff
@@ -77,22 +152,44 @@ export const PoolDetailLiquidity: React.FC = () => {
...
-              onClick={closeEpoch}
-              disabled={!pool}
-              loading={pool.epoch.isInSubmissionPeriod}
-              loadingMessage="Calculating..."
-            >
-              Close
-            </Button>
+            {isInChallengePeriod && (
+              <Text variant="label2">
+                Epoch is in challenge period until block {challengePeriodEnd.toString()} is finalized
```
offerijns, line 165: 1) I think we want to use the term calculation period rather than challenge period (but maybe check with Esther/Nico on this)
2) I think we should convert this to something like
```suggestion
                Epoch is in challenge period until {(challengePeriodEnd - currentBlock) * avgTimePerBlock / 60} min from now
```

sophialittlejohn: That's a good suggestion. I spoke with Esther about it and we've decided to remove it for now and tackle the count down in a separate ticket

centrifuge-app/src/components/EpochList.tsx
sophialittlejohn: turns out this is also already handled in cent-js (got lost in merge conflicts), I'll keep an eye out in case I see something like this again and will implement the method `toTokenBalance()` then 👍

offerijns (commented):
LGTM!

sophialittlejohn (approved):

PR #917, opened 2022-07-04
Centrifuge App: Fix pools refetching

onnovisser:
### Description

Checks for more events when refetching pools, so pools now also get refetching when the max reserve gets set, the epoch is closed, etc.

### Approvals

- [ ] Dev

### Impact

Centrifuge App


sophialittlejohn (approved):
💯 💪

PR #918, opened 2022-07-05
CentrifugeJS: Fix pools query not resolving when there are no pools

onnovisser:
### Description

Fixes a bug where the pools query never resolves, when there are no pools.

### Approvals

- [ ] Dev

### Impact

CentrifugeJS


sophialittlejohn (approved):
🤞✅

PR #926, opened 2022-07-08
CentrifugeJS: Refactor loan types

sophialittlejohn:
### Description

This pull request...

Since loans were separated on-chain into loans, activeLoans and closedLoans I updated our typings in cent-js to more closely resemble the chain. The main benefit however is that lots of our optional chaining and excessive property checking can be omitted thanks to the discriminated union (TS) 🎉 

I'm still looking for a better way to handle the different loan types within `loanInfo` (CreditLine, CreditLineWithMaturity and BulletLoan) since ts doesn't support nested discriminated unions...

Closes #<issue-number>

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product


onnovisser (approved):
💯

Issue #929, opened 2022-07-11
CentrifugeJS: Error handling for insufficent Native balance


sophialittlejohn:
1. return error from centrifuge-js if a user tries to make a tx with insufficient native token balance
2. add info message about insufficient balance to invest form and disable button

EstherCentrifuge:
works now, thanks!

PR #937, opened 2022-07-14
CentrifugeJS: Add error handling for insufficient native balances

sophialittlejohn:
### Description

This pull request... 
- stops the transaction before signing if the native balance is too low. It sends error message into toast (before `sendAndSign` is called on any tx, we make sure the native token balance is greater than the estimated fee)
- adds the possibility to get the tx fee from `useCentrifugeTransaction` so it can be displayed to the user. For now I've only implemented it in the invest form.

#929 

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product

### DIFF
--- a/centrifuge-app/src/utils/useCentrifugeTransaction.ts
+++ b/centrifuge-app/src/utils/useCentrifugeTransaction.ts
@@ -8,7 +8,7 @@ import { Transaction, useTransaction, useTransactions } from '../components/Tran
 import { useWeb3 } from '../components/Web3Provider'
 import { PalletError } from './errors'
 
-type TxOptions = Pick<TransactionOptions, 'propose'>
+type TxOptions = Pick<TransactionOptions, 'propose' | 'paymentInfo'>
 
 export function useCentrifugeTransaction<T extends Array<any>>(
   title: string,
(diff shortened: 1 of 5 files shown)


centrifuge-js/src/CentrifugeBase.ts
```diff
@@ -120,29 +125,59 @@ export class CentrifugeBase {
...
-      )
+
+      const $paymentInfo = submittable.paymentInfo(signingAddress)
+      const $balances = api.query.system.account(signingAddress)
+
+      if (options?.paymentInfo) {
+        return lastValueFrom(
+          $paymentInfo.pipe(
+            map((paymentInfo) => {
+              return paymentInfo.toJSON()
```
onnovisser, line 142: using a `Balance` might be nicer

centrifuge-app/src/utils/useCentrifugeTransaction.ts
```diff
@@ -108,6 +116,10 @@ export function useCentrifugeTransaction<T extends Array<any>>(
         updateTransaction(id, { status: 'failed', failedReason: e.message })
       })
     } else {
+      if (options?.paymentInfo) {
```
onnovisser, line 119: I'm not too sure about overloading `useCentrifugeTransaction` like this. I initially created a separate `useTransactionFeeEstimate` hook. Not sure if that one even works as I couldn't test it with the local chain at the time. But I do like the idea of at least having a hook where it's obvious that it's getting the payment info, even if it were to just abstract the paymentInfo function in this one.

onnovisser (approved):
💸

PR #939, opened 2022-07-14
chore: update asset ui page

### LINKED ISSUE #632: Remove "Actual investor capacity info" box for Maker pools on pool Management sheet
For Maker pools the info box

on the Management tab, "liquidity sub tab"
is not so helpful plus the calcs are partly off due to the Maker credit line.
For now, we should just remove it for pools with Maker integration.

### LINKED ISSUE #924: Update loan status and  asset detail page to display write offs for repaid assets
### What and why
- Tinlake has certain write-off groups on chain where overdue assets can be moved into to reduce their valuation in the NAV, e.g. an overdue asset with valuation of DAI 100, which is moved in a 10% write-off group will be valued with DAI 90 in the NAV
- Overdue assets are already being written-off for valuation on chain, but this is not reflected in the UI at all
- While showing all write-offs for ongoing loans in the UI will be difficult to build, we need to start showing at least write-offs of the residual of repaid loans on the asset details page (e.g. outstanding is 112, only 105 will be repaid in total, so the remaining 7 need to be written off)
- So we only show the write-off percentage for "closed" loans (see change in loan status below) but not if a there is a temporary write-off for an ongoing loan. This will be build next
- Once we are on implementing this, since we have much more asset data available in the sub-graph now, some of which can easily be added to the asset details page, we will combine this with minor update of the asset details page

### Changes to loan status
Background: 
- For now we can only show write-offs for repaid loans that were not repaid in full
- These are still on going with remaining outstanding debt on chain and would be moved in a write-off group with a write-off percentage of 100% (`Write-off group %`)
- When a loan is moved to this write-off group it is not automatically closed, so we need to catch this in the UI

What to change to `Loan Status`
- We need to include one check: If write-off % of write-off group (`Write-off group %`) = 100 % (careful, this is equivalent to a write-off group with write-off = 0 on chain, as on chain this is implemented as 0% of the value should be considered in NAV), then the loan status should be `Repaid` in the UI even if the on chain loan status is `Ongoing`
- Note: Please double-check, that in the Asset list, the amount shown in the `Amount` column is then the `total repaid`, which is the default for `Repaid` loans

#### Overview of changes
- There is a general change to the components and variables shown for all loan stati
- We add a few more variables we now have in the subgraph, such as `total repaid`
- With more variables now, the current status component will be split into "Status" and "Risk" 
- We add a `Write-off %` (see calculations below) to the "Risk" component only for `repaid` loans and loans and `ongoing `loans in a 100% write-off group (which are also displayed as `repaid`. For normal `ongoing`, `overdue` and `locked NFT` it won't be shown
- Note that we don't have extra mock-ups for all this, design of components etc all remain the same, we just add and re-order fields, will display the variables in a two column table below to resemble the current design

#### Changes to status component
- "Status" remains as title
- `Status pill` still shown as is it now
- Variables shown

| Name     | Format   | Name    | Format |
| -------- | -------- | ------- | -------- |
| Available for Financing | x,xxx.xx DAI | Financing Fee | x.xx %|
| Outstanding | x,xxx.xx DAI | Total Financed | x,xxx.xx DAI |
| Maturity Date | mmm dd, yyyy | Total repaid | x,xxx.xx DAI |
| Financed by | ETH address | [Empty] | [Empty] |

Notes:
New variables:
- Total financed
    - Can be grappbed from subgraph per asset: `Loan.borrowsAggregatedAmount`
- Total repaid
    - Can be grabbed from subgraph per asset: `Loan.repaysAggregatedAmount`

Moved variables:
- Risk group
    - Moved to risk component (see below)

#### Changes to "Risk" component
Overview
- This is a new component which combines existing variables with a few new ones
- Same design as status component, just don't show the status pill, so basically title and variables in two columns
- NOTE THAT the design varies for loans with status `NFT locked`, `ongoing`, `overdue x days` and "loan is in write-off group 100% state" (only in the latter case, we show the `write-off %``) 
- Title: "Risk"

##### Variables shown for loans in status `NFT locked`, `ongoing`, `overdue x days`:

| Name     | Format   | Name    | Format |
| -------- | -------- | ------- | -------- |
| Risk group | Integer | Applied risk adjustment | x.xx %|

Notes:
- Applied risk adjustment
    - Can be grapped from chain based on the risk group
    - We already grab this for the Portfolio component on the asset tab (Risk adjustments are per risk group, not per asset)
    - Needs tooltip: Same tooltip as shown for "Assumed risk adjustement" in the portfolio distribution componennt on the asset tab, so "This is the assumed risk adjustment applied to outstanding financings ..."
    - Note: Can we rename "Assumed risk adjustement" to "Applied risk adjustment" in the portfolio distribution componennt on the asset tab? 

##### Variables shown for loans in status `repaid` and for `ongoing`, `overdue x days` which are in a 100% write-off group

| Name     | Format   | Name    | Format |
| -------- | -------- | ------- | -------- |
| Risk group | Integer | Applied risk adjustment | x.xx %|
| Write-off % | x.xx %| [Empty] | [Empty] |

Notes
- Write-off %
    - Calculated as: 
`Write-off% = Debt at write-off / (Total borrowed + total debt accrued)` (with `Debt at Write-off` synonym/equal to `Write-off amount`)
    - We have both `total borrowed` and `total debt accrued`
    - We don't have the `Debt at write-off` but we have the `Current debt` (remember, the loan unfortunately continues to accrue debt while valued with 0 in a 100% write-off group) and if we know the `Time of write-off` we can re-calculate the `Debt at write-off`
    - To grab the `Time of write-off` we can 
        * Get list of events from PILE contract
        * Filter `ChangeRate` events for this `loan_id`
        * Get timestamp of last event
    - The formula to calculate `Debt at write-off` is then relatively simple: 
`Debt at write-off = Current debt / (fee^time since write-off in seconds)` so with `time since write-off in seconds` calculate as `now - Time of write-off`. As we move through time, current debt and time in seconds both grow and Debt at write-off will remain stable
    - Needs tooltip: "This indicates the percentage of the expected repayment that was not repaid and was written-off"
    - See examples calcs here: https://docs.google.com/spreadsheets/d/1KL0bSsk9GChy9mVM8S3mD-PnDCdweC0rH5WgYtzL-vo/edit#gid=1752100876

#### NFT data
- Can stay as it is

#### Finance/repay
- Can stay as it is

jpangelle:
### Description

This pull request

- makes changes to the asset page
- removes the `Alert` for pools on the management page

[Preview link](https://pr-939--dev-tinlake.netlify.app/pool/0x4b6ca198d257d755a5275648d471fe09931b764a/fortunafi-1/assets/asset?assetId=3)

Closes #632 
Closes #924 

### Approvals

- [x] Dev
- [ ] Product

### Screenshots

image

### DIFF
--- a/tinlake-ui/components/Loan/Data/index.tsx
+++ b/tinlake-ui/components/Loan/Data/index.tsx
@@ -54,12 +54,12 @@ const LoanData: React.FC<Props> = (props: Props) => {
         .filter((riskGroup) => riskGroup.ceilingRatio && !riskGroup.ceilingRatio.isZero())
         .find((riskGroup) => riskGroup.id === props.loan?.riskGroup)
     }
-  }, [poolData])
+  }, [poolData, props.loan?.riskGroup])
 
   const { showTransferCurrency } = useDebugFlags()
   const availableForFinancing = props.loan?.debt.isZero() ? props.loan?.principal || new BN(0) : new BN(0)
 
-  const { data: loanData } = useLoan(props.poolConfig.addresses.ROOT_CONTRACT, props.loan?.loanId as string)
+  const { data: loanData } = useLoan(props.poolConfig.addresses.ROOT_CONTRACT, Number(props.loan?.loanId))
 
   const appliedRiskAdjustment = React.useMemo(() => {
     if (riskGroup?.recoveryRatePD) {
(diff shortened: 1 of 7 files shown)


tinlake-ui/utils/useLoan.ts
```diff
@@ -0,0 +1,17 @@
+import { useQuery } from 'react-query'
```
jpangelle, line 1: new hook 🎉

tinlake-ui/containers/PoolManagement/Liquidity.tsx
```diff
@@ -6,7 +6,6 @@ import { Box, Button, Heading, Table, TableBody, TableCell, TableRow } from 'gro
 import * as React from 'react'
```
jpangelle, line 6: This is the work for #632 -- removing the `Alert`

tinlake-ui/components/Loan/Data/index.tsx
```diff
@@ -10,8 +10,11 @@ import { createTransaction, TransactionProps } from '../../../ducks/transactions
 import { addThousandsSeparators } from '../../../utils/addThousandsSeparators'
```
jpangelle, line 10: This file is the bulk of the work:

- Uses `usePool` to get data to compute the `riskGroup`.
- Uses `useLoan` to get data for the _Total financed_, _Total repaid_, and to calculate the `appliedRiskAdjustment`, 
- Makes edits to the table to break it up into two tables and add new rows

tinlake-ui/components/Scorecard/index.tsx
```diff
@@ -145,8 +145,8 @@ const Scorecard: React.FC<Props> = (props: Props) => {
                   </TableCell>
```
jpangelle, line 145: Requested verbiage changes

tinlake-ui/containers/PoolManagement/Risk.tsx
```diff
@@ -137,8 +137,8 @@ const Risk: React.FC<Props> = (props: Props) => {
               <TableCell size="22%">Max advance rate</TableCell>
```
jpangelle, line 137: Requested verbiage changes

tinlake-ui/public/static/tooltips.json
```diff
@@ -81,8 +81,8 @@
     "description": "Underlying every Tinlake pool is a \"risk scorecard\" with different risk groups. Each NFT is assigned to one of these risk groups to determine financing fee and assumed risk adjustment for valuation.",
```
jpangelle, line 81: Requested verbiage changes

tinlake-ui/containers/PoolManagement/Risk.tsx
```diff
@@ -137,8 +137,8 @@ const Risk: React.FC<Props> = (props: Props) => {
...
               <TableCell size="22%">Financing fee (APR)</TableCell>
               <TableCell size="22%">
-                <Tooltip id="assumedRiskAdjustment" underline>
-                  Assumed risk adjustment
+                <Tooltip id="appliedRiskAdjustment" underline>
+                  Applied risk adjustment
```
jpangelle, line 141: Requested verbiage changes

tinlake-ui/components/Scorecard/index.tsx
```diff
@@ -145,8 +145,8 @@ const Scorecard: React.FC<Props> = (props: Props) => {
...
                   <TableCell size="22%">Financing fee (APR)</TableCell>
                   <TableCell size="22%">
-                    <Tooltip id="assumedRiskAdjustment" underline>
-                      Assumed risk adjustment
+                    <Tooltip id="appliedRiskAdjustment" underline>
+                      Applied risk adjustment
```
jpangelle, line 149: Requested verbiage change

tinlake-ui/public/static/tooltips.json
```diff
@@ -81,8 +81,8 @@
...
     "link": "https://docs.centrifuge.io/learn/pool-valuation/"
   },
-  "assumedRiskAdjustment": {
-    "title": "Assumed risk adjustment",
+  "appliedRiskAdjustment": {
+    "title": "Applied risk adjustment",
```
jpangelle, line 85: Requested verbiage change

sophialittlejohn (commented):
Nice 🙌  code looks good to me! 

I noticed the same small UI things as Dennis (padding and some value never finish loading). Tested here https://pr-939--dev-tinlake.netlify.app/pool/0x4b6ca198d257d755a5275648d471fe09931b764a/fortunafi-1/assets/asset?assetId=3

tinlake-ui/utils/useLoan.ts
onnovisser: 🎉

tinlake-ui/utils/useLoan.ts
```diff
@@ -0,0 +1,17 @@
...
+import Apollo from '../services/apollo'
+
+export function useLoan(poolId: string, loanId: string) {
+  return useQuery(
+    ['loans'],
```
onnovisser, line 6: `poolId` should be in the key here, otherwise for other pools, you'll get the same loans, because the key won't have changed, so it grabs the cached value.

tinlake-ui/utils/useLoan.ts
```diff
@@ -0,0 +1,17 @@
...
+    async () => {
+      if (poolId && loanId) {
+        const { data: loansData } = await Apollo.getLoans(poolId)
+
+        const loanData = loansData.find((loan) => loan.loanId === loanId)
```
onnovisser, line 11: The `find` will have to be outside of the `useQuery`. Without having `loanId` in list of keys, `useQuery` will cache the result of this function and then serve the cached result for other loanIds. Having `loanId` in the keys would refetch all loans again when the loanId changed, so also not ideal. So best to have the `find` separate from the `useQuery` I think.

tinlake-ui/components/Loan/Data/index.tsx
```diff
@@ -52,98 +87,154 @@ const LoanData: React.FC<Props> = (props: Props) => {
...
-        </Box>
-      </Flex>
-      {props.loan?.ownerOf && props.loan?.borrower && showTransferCurrency && (
-        <ButtonGroup>
-          <Button label="Transfer currency from proxy" size="small" onClick={() => proxyTransfer()}></Button>
-        </ButtonGroup>
-      )}
-    </Stack>
+    <>
+      <Stack as={Card} gap="medium" p="medium" maxWidth={{ medium: 900 }}>
```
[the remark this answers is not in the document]
jpangelle: nice that fixed it!

tinlake-ui/utils/useLoan.ts
jpangelle: great point, adding it

tinlake-ui/utils/useLoan.ts
jpangelle: Looks like breaking the find out of the hook still results in the loan data being cached.

jpangelle: However, adding the `loanId` as a key fixes the issue, but yeah not ideal since it has to keep fetching for the loan data

onnovisser (approved):
🙌

sophialittlejohn (approved):
💯

PR #941, opened 2022-07-15
Centrifuge App: Add invest to pool page

onnovisser:
### Description

Adds the possibility to invest in each token to the pool detail page. Not including in this is the onboarding card and connecting to onboarding.

### Approvals

- [ ] Dev

### Screenshots

image

### Impact

Centrifuge App


centrifuge-app/src/pages/Pool/Overview/index.tsx
```diff
@@ -44,6 +155,43 @@ export const PoolDetailOverview: React.FC = () => {
     pageSummaryData.splice(2, 0, { label: <Tooltips type="age" />, value: getAge(pool.createdAt) })
   }
 
+  const tokens = pool?.tranches
+    .map((tranche) => {
+      const protection = tranche.minRiskBuffer?.toDecimal() ?? Dec(0)
+      const tvl = pool.value.toDecimal()
+      const totalReserve = pool.reserve.total.toDecimal() ?? Dec(0)
+      const maxReserve = pool.reserve.max.toDecimal() ?? Dec(0)
+      const capacityGivenMaxReserve = maxReserve.minus(totalReserve)
```
offerijns, line 164: Awesome that you started implementing this here! I assumed we were going to use the solver for this, but I think you are right that we can also implement it like this (similar to the Tinlake implementation). Two comments though:
- I think this should be moved to centrifuge.js, just adding a new `capacity` field to each tranche returned by `getPool()`.
- These calculations should take into account the current orders, e.g. once someone submits an investment order, this reduces the capacity. So `capacityGivenMaxReserve = maxReserve - totalReserve - outstanding Investment orders + outstanding redemption orders`. The more complicated calculation (but not impossible) is `capacityGivenProtection`: for each tranche, the new risk buffer assuming existing orders are executed, up to the limits, should be calculated.

Once this is moved to centrifuge.js, it would also be easy to add the capacities to the pool/token lists.

offerijns (commented):
This is looking really good! Took a quick look at the capacity calculations, cool that you implemented this, I left a few small suggestions.

centrifuge-app/src/utils/usePools.ts
```diff
@@ -40,13 +42,40 @@ export function useOrder(trancheId: string, address?: string) {
...
     {
-      enabled: !!address && !!pool,
+      enabled: !!address && !!pool && !!trancheId,
+    }
+  )
+
+  return result
+}
+
+export function usePendingCollectMulti(poolId: string, trancheIds?: string[], address?: string) {
```
sophialittlejohn, line 58: pendingCollectMulti might be useful in cent-js, something to keep an eye on

sophialittlejohn (approved):
Nice job 💪

PR #945, opened 2022-07-18
Centrifuge App: Add `notePreimage` option to create pool

sophialittlejohn:
### Description

This pull request...

Closes #<issue-number>

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product

### DIFF
--- a/centrifuge-app/src/pages/IssuerCreatePool/index.tsx
+++ b/centrifuge-app/src/pages/IssuerCreatePool/index.tsx
@@ -217,13 +217,18 @@ const CreatePoolForm: React.VFC = () => {
     setTimeout(() => setWaitingForStoredIssuer(false), 10000)
   }, [])
 
+  const txMessage = {
+    immediate: 'Create pool',
+    propose: 'Submit pool proposal',
+    notePreimage: 'Noting preimage (fast track)',
+  }
   const { execute: createPoolTx, isLoading: transactionIsPending } = useCentrifugeTransaction(
-    config.requiresPop ? 'Submit pool proposal' : 'Create pool',
+    txMessage[config.poolCreationType || 'immediate'],
     (cent) => cent.pools.createPool,
     {
       onSuccess: (args) => {
         const [, poolId] = args
-        if (!config.requiresPop) {
+        if (config.poolCreationType !== 'propose' && config.poolCreationType !== 'notePreimage') {
           history.push(`/issuer/${poolId}`)
         }
       },
@@ -341,7 +346,7 @@ const CreatePoolForm: React.VFC = () => {
           metadataHash,
           writeOffGroups,
         ],
-        { propose: config.requiresPop }
+        { createType: config.poolCreationType }
       )
 
       setSubmitting(false)
@@ -397,7 +402,7 @@ const CreatePoolForm: React.VFC = () => {
   )
 
   React.useEffect(() => {
-    if (config.requiresPop) {
+    if (config.poolCreationType === 'propose') {
       getProposeFee(form.values)
     }
   }, [form.values, getProposeFee])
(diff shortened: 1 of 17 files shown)


centrifuge-app/src/pages/IssuerCreatePool/index.tsx
```diff
@@ -217,13 +217,18 @@ const CreatePoolForm: React.VFC = () => {
     setTimeout(() => setWaitingForStoredIssuer(false), 10000)
   }, [])
 
+  const txMessage = {
+    immediate: 'Create pool',
+    propose: 'Submit pool proposal',
+    notePreimage: 'Noting preimage (fast track)',
```
offerijns, line 223: This isn't really fast-tracking it yet, it's just what enables the council to then fast track it afterwards, so I think we should change the copy a bit:

```suggestion
    notePreimage: 'Noting preimage',
```

centrifuge-js/src/modules/pools.ts
```diff
@@ -445,13 +445,19 @@ export function getPoolsModule(inst: CentrifugeBase) {
...
           const proposalSubmittable = api.tx.utility.batchAll([
             api.tx.democracy.notePreimage(submittable.method.toHex()),
             api.tx.democracy.propose(submittable.method.hash, api.consts.democracy.minimumDeposit),
           ])
           return inst.wrapSignAndSend(api, proposalSubmittable, options)
         }
+        if (options?.createType === 'notePreimage') {
+          const preimageSubmittable = api.tx.utility.batchAll([
+            api.tx.democracy.notePreimage(submittable.method.toHex()),
+          ])
```
offerijns, line 458: The batch tx is not needed in this case

```suggestion
          const preimageSubmittable = api.tx.democracy.notePreimage(submittable.method.toHex())
```

offerijns (approved):
Looks great 👍

centrifuge-app/src/pages/IssuerCreatePool/index.tsx
```diff
@@ -397,115 +407,140 @@ const CreatePoolForm: React.VFC = () => {
...
+    if (config.poolCreationType === 'notePreimage') {
+      const $events = cent
+        .getEvents()
+        .pipe(
+          filter(({ api, events }) => {
+            const event = events.find(({ event }) => api.events.democracy.PreimageNoted.is(event))
+            const parsedEvent = event?.toJSON() as any
+            // the events api returns a few events for the event PreimageNoted where the data looks different everytime
+            // when data is a tuple and the length is 3, it may be safe to extract the first value as the preimage hash
+            if (parsedEvent?.event?.data?.length === 3) {
```
sophialittlejohn, line 425: sadly, I couldn't find a better way to parse out the preimage hash coming in from the `PreimageNoted` event. Basically, this filter emitted more than one event for the condition  `api.events.democracy.PreimageNoted.is(event)` but not all of them had the preimage hash in their data. @offerijns do you know why and a better way to figure out which value is actually the preimage hash? For now this works but it feel very "hacky"

offerijns: @branan can you take a look at this?

branan: This is not the first report I've seen of weird event behavior. Unfortunately I've never been able to reproduce in my environment. I poked it a bit but only ever got a single event in my very basic tests of storing a preimage

offerijns: Since this feature isn't critical (worst case, we can find the hash on Subscan) for its use, I would say let's hope it works properly and just go ahead and merge this then.

PR #959, opened 2022-07-22
Centrifuge App: Show invest card when not whitelisted

onnovisser:
### Description

Show invest card in sidebar even when not whitelisted as a placeholder until we have the onboarding stuff.

### Approvals

- [ ] Dev


sophialittlejohn (approved):
🙌

PR #961, opened 2022-07-25
feat: add write off info

### LINKED ISSUE #960: Display write-off % of Write off group in UI
With https://github.com/centrifuge/apps/issues/924), we now can display additional asset details, e.g. also risk group and risk adjustment.

Unfortunately, we cannot display the actual write-off percentage of a partially repaid loan  BUT with the work made, we can at least
(i) indicate that a loan is "closed/repaid/finally written off" despite an outstanding on chain that keeps on growing (otherwise, if a loan is in a write-off group with a write-off percentage of 100% the loan will still always be ongoing with the outstanding growing)
(ii) and with little additional effort show the current write-off percentage for loans that are in a write-off group.

To implement both, the first step is to fetch whether the asset has been moved in a write-off group and check for the write-off % of this group (remember, it is diametrical on chain, so on chain "30% asset value percentage" is conceptually a 70% write-off percentage).
Then, 

(i) if write-off% of write off group = 100%
Change/Display loan status to `repaid`
Display outstanding = 0.
Add parameter `Applied write-off` to the risk component and display `100%` (if possible to display `Residual write-off` instead of 100% would be awesome, but assume it is a percentage field?) As we can't display the actual write-off % off the residuum, we can at least display the fact that it was written off. In combinatino with amount borrowed and repaid, this already gives a good approximation

(ii) if the write-off% of write-off group is > 0% but smaller tan 100%
Add parameter `Applied write-off` to the risk component and display the write-off %

(iii) if loan is not in a write-off group or write-off % is = 0%
Add parameter `Applied write-off` to the risk component and display `0%`

jpangelle:
### Description

This pull request adds another cell to display the "Applied write-off" to the risk table when viewing an asset. If the write-off % is 100%, the status will display "repaid" instead of "ongoing".

Closes #960 

### Approvals

- [x] Dev
- [x] Product

### Screenshots

image

### DIFF
--- a/tinlake-ui/components/Loan/List/index.tsx
+++ b/tinlake-ui/components/Loan/List/index.tsx
@@ -5,21 +5,19 @@ import BN from 'bn.js'
 import { DataTable, Text } from 'grommet'
 import { useRouter } from 'next/router'
 import * as React from 'react'
-import { useQuery } from 'react-query'
 import styled from 'styled-components'
 import NumberDisplay from '../../../components/NumberDisplay'
 import { Pool } from '../../../config'
 import { dateToYMD } from '../../../utils/date'
 import { hexToInt } from '../../../utils/etherscanLinkGenerator'
 import { saveAsCSV } from '../../../utils/export'
+import { useAssetListWriteOffStatus } from '../../../utils/useAssetListWriteOffStatus'
 import { useMedia } from '../../../utils/useMedia'
-import { calculateWriteOffPercentage } from '../../../utils/useWriteOffPercentage'
 import { ButtonGroup } from '../../ButtonGroup'
 import { Card } from '../../Card'
 import ChevronRight from '../../ChevronRight'
 import { useDebugFlags } from '../../DebugFlags'
 import { Box } from '../../Layout'
-import { useTinlake } from '../../TinlakeProvider'
 import LoanLabel from '../Label'
 
 interface Props {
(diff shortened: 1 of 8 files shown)


tinlake-ui/components/Loan/List/index.tsx
```diff
@@ -1,36 +1,57 @@
...
 const LoanList: React.FC<Props> = (props: Props) => {
   const router = useRouter()
   const { showExport } = useDebugFlags()
+  const tinlake = useTinlake()
+
+  const { isFetching: isFetchingLoansWithWriteOffStatus, data: loansWithWriteOffStatusData } = useQuery(
+    ['loansWithWriteOffStatus', props.loans],
+    async () => {
+      const loans: Loan[] = []
+      for (const loan of props.loans) {
```
onnovisser, line 40: I'd suggest doing a Promise.all here so it happens in parallel

tinlake-ui/utils/useWriteOffPercentage.ts
```diff
@@ -0,0 +1,32 @@
...
+  } catch {
+    return '0'
+  }
+}
+
+export function useWriteOffPercentage(poolId: string, loanId: number) {
+  const tinlake = useTinlake()
+
+  return useQuery(['writeOff', poolId, loanId], async () => calculateWriteOffPercentage(tinlake, loanId), {
+    enabled: !!poolId,
```
onnovisser, line 30: `&& !!loanId`

onnovisser (commented):
Nice, just left a couple comments

tinlake-ui/components/Loan/List/index.tsx
jpangelle: yeah refactored to use multicall, may want to give it another glance when u have a moment

tinlake-ui/utils/useWriteOffPercentage.ts
jpangelle: ^^

onnovisser (approved):
Nice 🙌

PR #967, opened 2022-07-26
Centrifuge App: Schemas WIP

onnovisser:
### Description

Initial work on schemas. Not connected to the node yet

### Approvals

- [ ] Dev

### Screenshots

image
image
image


centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
```diff
@@ -0,0 +1,324 @@
...
+  })
+
+  const formRef = React.useRef<HTMLFormElement>(null)
+  useFocusInvalidInput(form, formRef)
+
+  if (redirect) {
+    return <Redirect to={redirect} />
+  }
+
+  function getInput(attr: Schema['sections'][0]['attributes'][0]) {
```
sophialittlejohn, line 119: so I have a suggestion for this. If we map all of our input types directly to components then we can use something like this `createComponentSwitch`

```
const createComponentSwitch =
  (Components, type, fallback = null) =>
  (props) => {
    if (typeof props[type] !== 'undefined' && props[type] in Components) {
      return React.createElement(Components[props[type]], props);
    }

    if (fallback) {
      return React.createElement(fallback, props);
    }
    console.warn('component not defined');
    return null;
  };

const Switch = createComponentSwitch({
  'select': SchemaSelect,
  'currency': CurrencyInput,
});

const SchemaInput = (props) => {
  return (
    <Switch
      component={props.component}
      { ...props }
    />
  );
};
```
The reason I'm thinking of it this way is so that 
1. we can implicitly pass props; we might have to create implementations of our existing inputs just for the schema but at least we'd have a clear understanding of how the inputs are used in the schemas. And we have a clear overview of what types of fields are implemented and which ones are still necessary. So when the requirements change (which they will) we can easily decide to extend what we've got or create a whole new input type/implementation.  
2. we can save some nested ifs and provide an explicit fallback for each input if something was defined incorrectly. We'd have one identifier for each component so less ambiguity, which would also make documentation of the input types easier.

Anyway, let me know what you think. I've used this pattern in a couple of projects and found it nice to work with 😄

centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
```diff
@@ -0,0 +1,324 @@
...
+              />
+            </Grid>
+          </PageSection>
+        </Stack>
+      </Form>
+    </FormikProvider>
+  )
+}
+
+function truncate(txt: string, num: number) {
```
sophialittlejohn, line 315: this also exists like 20 times in our code already 🙈  util?

sophialittlejohn (commented):
What a great start 😍 I left a few suggestions 😄

centrifuge-app/src/pages/IssuerPool/Configuration/CreateSchema.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { FieldWithErrorMessage } from '../../../components/FieldWithErrorMessage'
+import { PageHeader } from '../../../components/PageHeader'
+import { PageWithSideBar } from '../../../components/PageWithSideBar'
+import { Schema } from '../../../types'
+import { useCentrifugeTransaction } from '../../../utils/useCentrifugeTransaction'
+import { usePrefetchMetadata } from '../../../utils/useMetadata'
+import { usePool, usePoolMetadata } from '../../../utils/usePools'
+import { pinPoolMetadata } from './pinPoolMetadata'
+
+const isValidJsonString = (str: string) => {
```
sophialittlejohn, line 14: this could probably also be a util

centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
onnovisser: hahaha yea will do 😅

centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
```diff
@@ -0,0 +1,324 @@
+import {
```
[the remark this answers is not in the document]
onnovisser: In code we call it loan basically everywhere else. it's just when we display it to the user that we call it asset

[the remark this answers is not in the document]
sophialittlejohn: makes sense, but this component in this file is called `CreateAsset`, so how about renaming it instead?

centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
onnovisser: I did think about doing something along these lines, but didn't because the need to create a bunch of schema-specific components to wrap the inputs. Maybe I'll give it a try, to see how tidy it looks

onnovisser: tried something

centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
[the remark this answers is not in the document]
onnovisser: yea makes sense, I'll rename the component

sophialittlejohn (approved):
Amazing!! Left a few comments but nothing blocking 💪

onnovisser:
made those changes you suggested

offerijns:
Awesome start! 🔥  A few comments after testing it a bit (probably many of them are already planned, just listing them regardless ;) ):
1. The table with asset templates on the Configuration page has Schema as the column name, and an Add schema button.
2. The code names everything schemas, I think we should rename this to asset templates for clarity.
3. The `Create asset` button should be primary.
4. The `Asset originator` dropdown should be removed.
5. The `issuer/create-asset` page should be a subpage of a specific pool.
6. The `schema` dropdown on the create asset page should be renamed to `Asset template`.
7. If `image` and or `description` in the asset template is `false`, then these should not even be shown on this page.
8. The `Asset type` dropdown should be added to the top, and a Pricing section needs to be added.

PR #970, opened 2022-07-26
fix: calculate new debt ceiling max

### LINKED ISSUE #969: Add buffer to setting locked credit line
Setting the `Locked Credit Line` on the Pool Management tab is currently limited by the minimum of `Max Locked Credit Line given TIN` and `Debt Ceiling`

This leads to issuers being able to set the `Locked Creditline `exactly to the Debt Ceiling which may introduce issues on smart contract level in certain epoch situations/ interest accruing in between etc. An easy fix is to always leave a buffer of 100k to the Debt ceiling forced through the UI.
So the limit should always be given as min(`Max Locked Credit Line given TIN`,`Debt Ceiling-100k`). This should impact the limit displayed below the line, plus the maximum number that is accepted as input for the transaction (e.g. if the input is above this number, input is in "error state" 
nd 
and `Update` button should be disabled (which it currently isn't, so it allows the user to run into a failed transaction)

jpangelle:
### Description

This pull request refactors the Locked Credit Line to the minimum of 1. Max Locked Credit Line given TIN and 2. Debt Ceiling - 100,000.

Closes #969 

### Approvals

- [x] Dev
- [x] Designer

### DIFF
--- a/tinlake-ui/containers/PoolManagement/Liquidity.tsx
+++ b/tinlake-ui/containers/PoolManagement/Liquidity.tsx
@@ -43,6 +43,10 @@ const Liquidity: React.FC<Props> = (props: Props) => {
   const mat = poolData?.maker?.mat
 
   const debtCeiling = (poolData?.maker?.line || new BN('0')).div(new BN(10).pow(new BN(45 - 18)))
+
+  // debt ceiling minus 100,000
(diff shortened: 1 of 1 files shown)


sophialittlejohn (approved):
I can't test the feature but code looks good 🙌

jpangelle:
> I can't test the feature but code looks good 🙌

Dennis was able to test it! Thx Soph 👍

PR #972, opened 2022-07-27
feat: add warning if cfg balance is zero

### LINKED ISSUE #971: Display warning that CFG is needed to link address
With a recent chain upgrade the transaction to link an CFG account to an investor ETH account to claim CFG now also costs (minimal) CFG. This means investors would already need CFG on their account to claim CFG. They'd need to get this CFG eg from exchanges or similar. We need to build at least a warning for this. So, quick fix, on the linking screen

Please add `Make sure this account is funded with CFG to cover transaction costs for the linking transaction. CFG is available on many exchanges.` in the warning message field after `Make sure ....cannot be undone`

Nice to have if you wanna build sth fancy: Fetch whether the selected account is holding any CFG and only display this message if it doesn't: `This account does not hold any CFG to cover transaction costs for the linking transaction. CFG is available on many exchanges.`

jpangelle:
### Description

This pull request adds a warning alert if the user's selected account balance is less than or equal to `0` when linking a wallet.

Closes #971 

### Approvals

- [x] Dev
- [x] Product

### Screenshots

**For account with balance above zero:**
Screen Shot 2022-07-26 at 7 11 04 PM

**For account with balance zero or below:**
Screen Shot 2022-07-26 at 7 10 56 PM

### DIFF
--- a/tinlake-ui/containers/SetCentAccount/index.tsx
+++ b/tinlake-ui/containers/SetCentAccount/index.tsx
@@ -1,4 +1,5 @@
 import Tinlake from '@centrifuge/tinlake-js'
+import BN from 'bn.js'
 import { ethers } from 'ethers'
 import { Box, Button, Select } from 'grommet'
 import { CircleAlert } from 'grommet-icons'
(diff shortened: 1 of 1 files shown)


onnovisser (approved):
`useQuery` might make this a bit nicer, but this works too 👍

denniswell:
Works, with my test accounts, thanks @jpangelle

jpangelle:
> `useQuery` might make this a bit nicer, but this works too 👍

good call, refactored.

Issue #951, opened 2022-07-19
CentrifugeJS: Integrate pool config saving/retrieving


sophialittlejohn:
To prepare for the node integration we need to move IPFS pinning and metadata fetching into centrifuge-js. 

> For context: When the node is ready to develop against, we will include a node endpoint and a schema map in the pool configs. Public pool information will be stored on IPFS and private information on the node

> Onno: We currently have multiple lambdas for IPFS stuff that are quite distinct and can do multiple things, like pinning an image and json when creating collection metadata. maybe instead of centrifuge.js needing to know how those lambdas are structured, we could pass generic pinFileToIpfs and pinJsonToIpfs callbacks to Centrifuge.js as config. 
```{ pinFileToIpfs: async (file: File) => ipfsHash }```

PR #973, opened 2022-07-27
CentrifugeJS: Integrate metadata pinning

sophialittlejohn:
### Description

This pull request...

- #951 

Moves metadata pinning to cent-js

### Approvals

- [ ] Dev


centrifuge-app/lambda/pinFile.js
```diff
@@ -24,7 +24,6 @@ const ipfsHashToURI = (hash) => `ipfs://ipfs/${hash}`
 const handler = async (event) => {
   try {
     const { fileDataUri, fileName } = JSON.parse(event.body)
```
onnovisser, line 26: wish we could get rid of the `fileName` somehow. it's only used to store a temp file. could we just randomly generate it?

centrifuge-js/src/CentrifugeBase.ts
```diff
@@ -39,6 +39,7 @@ export type Config = {
   centrifugeSubqueryUrl: string
   altairSubqueryUrl: string
   metadataHost: string
+  pinFile?: (values: any) => Promise<{ uri: string; ipfsHash: string }>
```
onnovisser, line 42: could `values` just be the file and have it be type File? Or string I guess as it's now dataUri

centrifuge-js/src/modules/metadata.ts
```diff
@@ -10,6 +10,45 @@ export function getMetadataModule(inst: Centrifuge) {
     return inst.getMetadataObservable<T>(url)
   }
 
+  function pinFile(metadata: {
+    fileDataUri?: string
+    fileName?: string
+  }): Observable<{ uri: string; ipfsHash: string }> {
```
onnovisser, line 16: wasn't this going to return just a string and we check if it's just an ipfs hash or a proper uri already?

sophialittlejohn: I've still left it returning an object with just the URI and in cent-js were parsing it into the full URI

EstherCentrifuge:
if this is deployed, can we close it please? @sophialittlejohn

sophialittlejohn:
@EstherCentrifuge that's just a PR deployment 😄 I'll close the PR and the ticket when I merge it into main (that's also when you'll be able to test it on demo and dev)

PR #974, opened 2022-07-27
Centrifuge App: Fix pool proposal

onnovisser:
### Description

Quick fix to get the pool proposal past democracy, logs the metadata hash and write-off groups, for doing those transactions after proposal passes.

### Approvals

- [ ] Dev


sophialittlejohn (approved):
I guess this will do 😢

Issue #940, opened 2022-07-15
Provide system feedback for epoch status


armandomistral:
## Goal
Centrifuge app shows the current state of the epoch, so users can see the progress (of the epoch duration), and whether orders are executable.

The epoch status should allow users to decide whether they want to wait for the epoch to close automatically, or to close it manually.

## Specification
[Epoch status - Centrifuge App V1
](https://centrifuge.hackmd.io/UkgOXautQI2E3pG7idKcsQ?view)

EstherCentrifuge:
the feedback looks good! 
I got some none solvable scenario's but the feedback given is good so i'll pass this.

PR #1022, opened 2022-08-24
Centrifuge App: Epoch status UI

sophialittlejohn:
### Description

This pull request...
- #940 

Closes #<issue-number>

### Approvals

- [ ] Dev


centrifuge-app/src/components/LiquidityProvider.tsx
```diff
@@ -0,0 +1,122 @@
...
+    }
+  )
+
+  const submitSolution = async () => {
+    if (!pool) return
+    submitSolutionTx([pool.id, true])
+  }
+
+  React.useEffect(() => {
+    submitSolution()
```
onnovisser, line 47: this confused me when I first saw it, because I didn't realize the tx had a dryRun option that doesn't actually do the tx. maybe a nicer way is to use a `useQuery` and get the `firstValueFrom` the `cent.pools.submitSolution`

onnovisser: also the stuff that's in `useMemo` could be in the query callback. this could then all be just in the `useLiquidity` hook to avoid having the Provider

onnovisser: the `useLiquity` hook might then take a pool id as param so it doesn't have to assume where that comes from

centrifuge-app/src/utils/useLiquidity.tsx
```diff
@@ -0,0 +1,98 @@
...
+export const useLiquidity = (poolId: string) => {
+  const pool = usePool(poolId)
+  const cent = useCentrifuge()
+  const { data: metadata } = usePoolMetadata(pool)
+  const [investments, setInvestments] = React.useState<LiquidityTableRow[]>()
+  const [redemptions, setRedemptions] = React.useState<LiquidityTableRow[]>()
+
+  const $source = cent.pools.submitSolution([poolId], { dryRun: true })
+
+  const { refetch } = useQuery(['solution', { poolId }], () => ($source ? firstValueFrom($source) : null), {
```
onnovisser, line 28: couldn't it just directly be `firstValueFrom(cent.pools.submitSolution([poolId], { dryRun: true }))`. I don't see how $source could be falsey

centrifuge-app/src/utils/useLiquidity.tsx
```diff
@@ -0,0 +1,98 @@
...
+          locked: token.outstandingRedeemOrders?.toDecimal() || new Decimal(0),
+          executing: solution?.tranches?.[index]?.redeem.amount.toDecimal() || new Decimal(0),
+          executingPercentage: solution?.tranches?.[index]?.redeem.perquintill || Perquintill.fromPercent(0),
+        }
+      })
+      setRedemptions(redemptionsRow)
+    },
+  })
+
+  React.useEffect(() => {
```
onnovisser, line 55: no need for this, useQuery will auto refetch when poolId changes

centrifuge-app/src/utils/useBlock.ts
```diff
@@ -0,0 +1,25 @@
...
+import React from 'react'
+import { useCentrifuge } from '../components/CentrifugeProvider'
+
+export const useBlock = () => {
+  const cent = useCentrifuge()
+  const [block, setBlock] = React.useState<Block>()
+
+  const $source = cent.getBlocks()
+
+  React.useEffect(() => {
```
onnovisser, line 11: `useCentrifugeQuery` would make sure the Observable is shared in case there's multiple uses of `useBlock`

onnovisser (dismissed):
Looking good!

centrifuge-app/src/utils/useLiquidity.tsx
sophialittlejohn: thanks! turns out I hadn't set good query keys, so the query wasn't going stale when new orders came in

onnovisser (approved):

PR #1036, opened 2022-09-07
Centrifuge App: Node integration for asset creation

onnovisser:
### Description

Integrates creating and minting documents on the node with the create asset flow

### Approvals

- [ ] Dev

### Screenshots

image

image

image

### Impact

CentrifugeJS, Centrifuge App


centrifuge-js/src/modules/pools.ts
```diff
@@ -1666,6 +1666,20 @@ export function getPoolsModule(inst: Centrifuge) {
     )
   }
 
+  function getNftDocumentId(args: [collectionId: string, nftId: string]) {
```
sophialittlejohn, line 1669: it seems this would fit better in the `nft` module

centrifuge-js/src/modules/pod.ts
```diff
@@ -0,0 +1,195 @@
...
+        Authorization: `Bearer ${token}`,
+      },
+    })
+      .then((res) => {
+        if (res.ok) return res
+        return res.json().then((res) => {
+          throw new Error(res.message)
+        })
+      })
+      .then((res) => res.json())
```
sophialittlejohn, line 101: I'd suggest a catch clause here just in case

sophialittlejohn (commented):
Nice job 💪  I guess it didn't deploy because of the conflicts, sorry about those!

One suggestion: minimal documentation, even just a link to the pod repo would be super useful

centrifuge-js/src/modules/pod.ts
onnovisser: In case the fetch promise throws, the `callPod` should still throw, I think. So having a `catch(err => { throw new Error(err))` wouldn't make sense to me

centrifuge-js/src/modules/pools.ts
onnovisser: yea it's a bit in-between, but will move

centrifuge-app/src/utils/useNFTs.ts
```diff
@@ -25,18 +26,34 @@ export function useNFT(collectionId?: string, nftId?: string, suspense = true) {
   return result
 }
 
-export function useLoanNft(poolId?: string, loanId?: string) {
+export function useLoanCollectionId(poolId?: string) {
```
sophialittlejohn, line 29: I'm not sure I fully understand this since collections don't belong to pools. NFTs belong to collections and can be "assigned" to pools, right? So technically it should be possible for a pool to have multiple collection ids. but this hook only handles one (which is why we passed the `loanId` previously)

centrifuge-app/src/utils/usePermissions.ts
```diff
@@ -11,20 +11,27 @@ export function usePermissions(address?: string) {
   return result
 }
 
-export function useCanBorrow(poolId: string, assetId: string) {
+export function useCanBorrow(poolId: string) {
```
sophialittlejohn, line 14: is this hook specifically for borrowing against NFTs? and the second one (`useCanBorrowAsset`) for pod documents? I think we could improve the naming a bit

centrifuge-js/src/CentrifugeBase.ts
```diff
@@ -111,7 +111,7 @@ export class CentrifugeBase {
   }
 
   getChainId() {
-    return this.config.network === 'centrifuge' ? 36 : 136
+    return 136
```
sophialittlejohn, line 114: should this be reverted? what ID does altair run on?

centrifuge-js/src/modules/pod.ts
```diff
@@ -0,0 +1,240 @@
...
+  response: JobResponse
+  constructor(jobResponse: JobResponse) {
+    super('Job failed')
+    this.response = jobResponse
+  }
+}
+
+/**
+ * Module to interact with the Centrifuge P2P Node for sharing private documents
+ * @see https://github.com/centrifuge/go-centrifuge
```
sophialittlejohn, line 89: ❤️

sophialittlejohn (commented):
It's coming along nicely!!

I do have a suggestion :upside_down_face:  its a bit big though… 

how would you feel about splitting `signAndSend` into 2 parts: `signTx` and `sendTx`? What would be really cool is if they’re chainable so in most cases we call `submittable.signTx().sendTx()`? 

There are 2 main benefits I see in this:
1. our code becomes shorter and easier to maintain (let’s face it, some of these function in cent-js have grown waaaay bigger than we want)
2. it will give us more fine grain control for future cases as well

centrifuge-app/src/utils/usePermissions.ts
onnovisser: `useCanBorrow` is returns whether an account can borrow from a pool in principle. `useCanBorrowAsset` returns whether an account can borrow a specific asset from a pool, so whether they have the borrow permission and own the asset

centrifuge-app/src/utils/useNFTs.ts
onnovisser: Each pool has 2 collections (currently), the loan collection which holds the loan NFTs and the collateral collection which holds the collateral (documents/art) NFTs. When you create a loan you lock the collateral NFT and mint a loan NFT. There could be more than one collateral collection and in the future there will be when there's multiple issuers to a single pool. At that point we'll have to add references to the collateral collections to the pool config. for now, when initializing the pool we also create 1 collateral collection, the ID of which is the loan collection id + 1. This way we can easily derive the collateral collection id when needing to mint document NFTs without having to store it on chian or look it up elsewhere.

centrifuge-js/src/CentrifugeBase.ts
onnovisser: yes reverted it for now, but I guess better would be to change this to `api.registry.chainSS58` so it's actually taking it from the connected node. the only annoying thing about that is that it then `formatAddress` can't be synchronous anymore

onnovisser:
> It's coming along nicely!!
> 
> I do have a suggestion 🙃 its a bit big though…
> 
> how would you feel about splitting `signAndSend` into 2 parts: `signTx` and `sendTx`? What would be really cool is if they’re chainable so in most cases we call `submittable.signTx().sendTx()`?
> 
> There are 2 main benefits I see in this:
> 
>     1. our code becomes shorter and easier to maintain (let’s face it, some of these function in cent-js have grown waaaay bigger than we want)
> 
>     2. it will give us more fine grain control for future cases as well

yea agree, the `wrapSignAndSign` most of all has gotten a bit out of hand. I'd like to hold of on separating the sign and send though until we have at least another use-case for it.

centrifuge-app/src/utils/usePermissions.ts
sophialittlejohn: how about a comment explaining the difference? Just trying to make it easier to distinguish

PR #1050, opened 2022-09-20
Updates .nvmrc

Hornebom:
### Description
- node v14.15.1 does not play along with yarn v2.4.3 (as stated in .yarnyml)
- solved by using node v16

### Approvals
- [x] Dev


jpangelle:
Were you able to confirm if all of the apps in the monorepo still run fine with this change?

Hornebom:
@jpangelle I could not get anything up and running with the currently specified version of `v14.15.1`. Rest of the team works on `v16` afaik. 
How would you approach such testing?

onnovisser (approved):
this has no impact on builds as it's only for NVM users. Don't see any risk merging this, if this is what was needed to install locally

jpangelle (approved):

PR #1051, opened 2022-09-20
build fabric before start

Hornebom:
### Description
- makes sure that fabric is build before start script is executed

The application will fail to run if fabric is not build once. I ran into that when starting the first time. Hope this change helps when onboarding new devs.

### Approvals
- [ ] Dev


onnovisser (approved):
💯

PR #1053, opened 2022-09-21
removes not used .env.*  references

Hornebom:
### Description
Removes all not used references to .env files. These were confusing to me, as we only use one single .env file. .env is ignored already on root level of the mono repo, but leaving it in place to make sure it will never end up getting pushed.

### Approvals
- [ ] Dev


onnovisser (approved):
👍

sophialittlejohn (approved):

PR #1061, opened 2022-10-03
Centrifuge App: Fix POD Auth

onnovisser:
### Description

Use the correct proxy type when creating the token to log into the POD.

### Approvals

- [ ] Dev


sophialittlejohn (approved):
💯

PR #1123, opened 2023-01-06
Fix packages

onnovisser:
### Description

Centrifuge App dev server wasn't working because of package changes in a recent commit

### Approvals

- [ ] Dev


sophialittlejohn (approved):
Thanks for fixing! LGTM 🙌

PR #1128, opened 2023-01-10
OnboardingAPI: Update endpoints: (`/verifyBusiness`, `/confirmOwners`, `/getUser`)

sophialittlejohn:
### Description

This pull request...
- remodels data for firestore (one user collection and one business collection with wallet address as common key)
- adds endpoints `/getUser` and `/getBusiness` (even though getUser returns the business as well)
- refactors some of the request body key names (mostly just removed `company` and `business` prefix)
- renames `businessVerification` --> `verifyBusiness`
- renames `businessVerificationConfirm` --> `confirmOwners`

### Approvals

- [x] Dev


jpangelle (approved):
looks great! and kudos for staying on top of documentation 😍

onboarding-api/README.md
```diff
@@ -14,28 +14,71 @@
...
+      {
+        "investorType": "entity",
+        "poolId": "123abc",
+        "trancheId": "0x123456"
+      }
+    ],
+    "steps": [
+      {
+        "step": "InvestorType",
+        "completed": true // set to true in request
```
jpangelle, line 53: since comments are invalid in json, this formats weirdly. might be better to just make the language as `js`

onboarding-api/src/controllers/user/createUser.ts
```diff
@@ -0,0 +1,44 @@
+import { Request, Response } from 'express'
```
jpangelle, line 1: We don't need this endpoint right, since the user will implicitly created in the `/verifyBusiness` endpoint?

sophialittlejohn: I forgot to remove it, thanks 😄

onboarding-api/src/controllers/kyb/getBusiness.ts
```diff
@@ -0,0 +1,17 @@
+import { Request, Response } from 'express'
+import { Business, userCollection } from '../../database'
+import { HttpsError } from '../../utils/httpsError'
+
+export const getBusinessController = async (req: Request, res: Response) => {
```
jpangelle, line 5: I don't think we need this endpoint either since the business information lives within the user object.

Issue #1109, opened 2022-12-09
Centrifuge Pools bugs


annamehr:
Here are the bugs that were found after testing Pools on dev.app.cntrfg.com. 

- [x] Error creating pool - invalid json - user added a normal description with hyphens but text was not accepted as valid json format (from 24:49 https://drive.google.com/file/d/1BsQtbPLtMDdFmzj3vTc_iSRinvYHLcoK/view) 

- [x]  No decimals allowed for fixed interest rate for tranches (only 7 or 8%) -> user should be able to choose interest rate with two decimals

PR #1134, opened 2023-01-13
CentrifugeApp: Prepare for gcloud migration

sophialittlejohn:
### Description

This pull request...

Closes #<issue-number>

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product


centrifuge-app/functions/src/pinFile.js
```diff
@@ -37,37 +37,32 @@ const dataUriToReadStream = (uri) => {
 
 const ipfsHashToURI = (hash) => `ipfs://ipfs/${hash}`
 
-const handler = async (event) => {
+const handler = async (req, res) => {
```
onnovisser, line 40: I think we should add an endpoint to directly pin a JSON string. The `jsonToBase64` we have in CentJS doesn't work when characters are used that are not in the Latin-1 character set, like certain dashes or emoji (See #1109).

onnovisser: I would still like to get rid of all the base64 encoding at some point and just upload files/json

PR #1136, opened 2023-01-13
Feat animated Collapsible

Hornebom:
### Description
- Creates animated fabric Collapsible component (if support for Web Animation Api and falls back to non animated variant)
- refactors InteractiveCard (fabric) and PageSection (app) to use Collapsible

### Approvals
- [ ] Dev


fabric/src/components/Collapsible/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const root = React.useRef<HTMLDivElement>(null)
+  const [enhanced, setEnhanced] = React.useState(false)
+  let animation: Animation | undefined = undefined
+
+  React.useEffect(() => {
+    setEnhanced(root?.current?.animate !== undefined)
+  }, [root])
+
+  React.useEffect(() => {
+    if (enhanced) {
```
onnovisser, line 32: can just be `root?.current?.animate !== undefined`. no need for the extra useEffect and useState

fabric/src/components/Collapsible/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+      animate()
+    }
+  }, [open, enhanced])
+
+  function animate() {
+    if (animation) {
```
onnovisser, line 38: `animation` should use a `useRef`. Otherwise this will always be `undefined`, as that's its initial value each time this component renders.

fabric/src/components/Collapsible/index.tsx
Hornebom: 👌  refactored

onnovisser (approved):

PR #1138, opened 2023-01-16
Centrifuge App: Fix create pool

onnovisser:
### Description

Fixed a couple of bugs on the Create Pool page

- Fixes converting pool metadata to base64, could fail when using special characters
- Fixes the redirect after creating a pool
- Fixes the issuer name and description fields being greyed out even though it looks like it finished loading stored values for the same address
- Fixes tooltip underline, when the text spans multiple lines

#1109 

### Approvals

- [ ] Dev

### DIFF
--- a/centrifuge-app/src/pages/IssuerCreatePool/index.tsx
+++ b/centrifuge-app/src/pages/IssuerCreatePool/index.tsx
@@ -14,6 +14,7 @@ import {
 } from '@centrifuge/fabric'
 import { Field, FieldProps, Form, FormikErrors, FormikProvider, setIn, useFormik } from 'formik'
 import * as React from 'react'
+import { useQueryClient } from 'react-query'
 import { useHistory } from 'react-router'
 import { filter, lastValueFrom } from 'rxjs'
 import { useCentrifuge } from '../../components/CentrifugeProvider'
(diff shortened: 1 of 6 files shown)


centrifuge-app/src/pages/IssuerCreatePool/index.tsx
```diff
@@ -150,6 +155,20 @@ const CreatePoolForm: React.VFC = () => {
...
+  React.useEffect(() => {
+    if (storedIssuer) setWaitingForStoredIssuer(false)
+  }, [storedIssuer])
+
+  React.useEffect(() => {
+    if (createdPoolId && pools?.find((p) => p.id === createdPoolId)) {
+      // Redirecting only when we find the newly created pool in the data from usePools
+      // Otherwise the Issue Overview page will throw an error when it can't find the pool
+      // It can take a second for the new data to come in after creating the pool
+      history.push(`/issuer/${createdPoolId}`)
```
Hornebom, line 167: Does the ui reflect/communicate this to the user? Using a spinner or something comparable?

Hornebom (approved):

centrifuge-app/src/pages/IssuerCreatePool/index.tsx
onnovisser: No, that would be a nice improvement. in production the redirect doesn't ever happen though, as the create pool has to pass governance first

onnovisser: the redirect is more of a convenience for the dev and demo envs

PR #1140, opened 2023-01-17
Adds feature flag to hide newly created pools

Hornebom:
### Description
- by default, any newly created pool is unlisted (listed: false)
- for the admin of the pool, it does also show up on the Investments tab, but not for anyone else
- don’t show the option to mark a pool as listed except through a feature flag

Closes 1118

### Approvals
- [ ] Dev

### DIFF
--- a/centrifuge-app/src/pages/Pools.tsx
+++ b/centrifuge-app/src/pages/Pools.tsx
@@ -11,7 +11,9 @@ import { Tooltips } from '../components/Tooltips'
 import { config } from '../config'
 import { Dec } from '../utils/Decimal'
 import { formatBalance } from '../utils/formatting'
+import { useAddress } from '../utils/useAddress'
 import { useMetadataMulti } from '../utils/useMetadata'
+import { usePermissions } from '../utils/usePermissions'
 import { usePools } from '../utils/usePools'
 
 export const PoolsPage: React.FC = () => {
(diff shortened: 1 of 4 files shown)


centrifuge-app/src/pages/Pools.tsx
```diff
@@ -24,18 +26,21 @@ export const PoolsPage: React.FC = () => {
...
+  const address = useAddress()
+  const permissions = usePermissions(address)
 
   const poolMetas = useMetadataMulti<PoolMetadata>(pools?.map((p) => p.metadata) ?? [])
-
+  console.log('poolMetas', poolMetas)
   const [listedPools, listedTokens] = React.useMemo(
     () => {
-      const listedPools = pools?.filter((_, i) => poolMetas[i]?.data?.pool?.listed)
+      const poolAdmins = pools?.map(({ id }) => permissions?.pools[id]?.roles.includes('PoolAdmin')) ?? []
```
offerijns, line 36: I just realized, we ideally would also show the pool for accounts with other roles in the pool. If possible, really any role on the pool should mean you see the pool. E.g. if you're an investor already in the pool, you definitely should see the pool (otherwise you can't redeem your investment). Also if you're a borrower, ...

centrifuge-app/src/pages/Pools.tsx
```diff
@@ -24,18 +26,21 @@ export const PoolsPage: React.FC = () => {
 
 const Pools: React.FC = () => {
   const pools = usePools()
+  const address = useAddress()
+  const permissions = usePermissions(address)
 
   const poolMetas = useMetadataMulti<PoolMetadata>(pools?.map((p) => p.metadata) ?? [])
-
+  console.log('poolMetas', poolMetas)
```
offerijns, line 33:
```suggestion
```

onnovisser:
Btw, one thing that I think was missing from the ticket is that this should only be done for the demo env

offerijns:
> Btw, one thing that I think was missing from the ticket is that this should only be done for the demo env

Good point! I guess we should add an env var `DEFAULT_UNLIST_NEW_POOLS` or something that is `true` for demo and `false` for everything else.

centrifuge-app/src/pages/Pools.tsx
Hornebom: Refactored to accept all addresses with permissions for the current pool

centrifuge-app/src/pages/Pools.tsx
Hornebom: 👌  removed

centrifuge-app/src/pages/IssuerCreatePool/index.tsx
```diff
@@ -108,6 +108,7 @@ const initialValues: CreatePoolValues = {
   epochHours: 23, // in hours
   epochMinutes: 50, // in minutes
   podEndpoint: config.defaultPodUrl ?? '',
+  listed: !window.location.hostname.includes('demo'),
```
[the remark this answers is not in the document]
Hornebom: I totally agree on that one and taking it a bit further 👌 

Refactored to use `REACT_APP_DEFAULT_UNLIST_NEW_POOLS` to control default listing and `REACT_APP_IS_DEMO` to define demo environment

.github/workflows/centrifuge-app-build-and-deploy.yml
```diff
@@ -38,6 +38,14 @@ on:
...
         required: true
         type: string
+      default-unlist-new-pools:
+        required: false
+        type: boolean
+        default: false
+      is-demo:
+        required: false
+        type: boolean
+        default: false
```
sophialittlejohn, line 48: Is it possible to use just one env variable for this? so pools are by default unlisted in demo?

sophialittlejohn (approved):
LGTM 💯

offerijns (approved):
LGTM!

PR #1176, opened 2023-02-01
Wallet in header

Hornebom:
### Description
- refactor PageWithSidebar to keep Wallet in header

### Approvals

- [ ] Dev
- [ ] Designer


onnovisser (approved):

annamehr:
Hi @Hornebom, I think the changes you made are part of a bigger redesign and not something we can fix now. As we need to take into consideration various components of the app, I'd suggest to leave this for now and once we have a FT designer onboard we can start tackling this bigger design challenge.

Hornebom:
Solved within 1473

PR #1177, opened 2023-02-01
Refactor to use native Select

Hornebom:
### Description
Refactor fabric `<Select/>` to use native `<select>`

### Approvals
- [ ] Dev

### DIFF
--- a/centrifuge-app/src/pages/Loan/PricingForm.tsx
+++ b/centrifuge-app/src/pages/Loan/PricingForm.tsx
@@ -149,8 +149,10 @@ export const PricingForm: React.VFC<{ loan: LoanType; pool: Pool }> = ({ loan, p
             <Field name="loanType">
               {({ field, meta, form }: FieldProps) => (
                 <Select
+                  id="select-asset-type"
+                  name="asset-type"
                   label="Asset type"
-                  onSelect={(v) => form.setFieldValue('loanType', v, false)}
+                  onChange={(event) => form.setFieldValue('loanType', event.target.value, false)}
                   onBlur={field.onBlur}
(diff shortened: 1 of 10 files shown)


fabric/src/components/Select/index.tsx
```diff
@@ -1,44 +1,22 @@
...
-  errorMessage?: string
-}
-
 export type SelectOptionItem = {
-  label: string | React.ReactElement
+  label: string
   value: string
 }
 
 export type SelectProps = {
```
onnovisser, line 16: I would suggest adding `React.InputHTMLAttributes<HTMLSelectElement>` and spreading extra props onto the select element. then we can remove `onBlur/value/id/name` and use the native `onChange` instead of the onSelect

Hornebom: Good point 👌

centrifuge-app/src/pages/Loan/PricingForm.tsx
```diff
@@ -167,8 +168,9 @@ export const PricingForm: React.VFC<{ loan: LoanType; pool: Pool }> = ({ loan, p
           <Field name="riskGroup" validate={required()}>
             {({ field, meta, form }: FieldProps) => (
               <Select
+                name="risk-group"
```
onnovisser, line 171: should match Formik field name `riskGroup`

Hornebom: All refactored 👌

onnovisser (approved):

sophialittlejohn (approved):
Not sure why this change but LGTM 👍

PR #1182, opened 2023-02-07
Centrifuge App: Wallet dialog

onnovisser:
### Description

Adds the possibility to connect EVM wallets and adds a wallet dialog

### Approvals

- [ ] Dev


centrifuge-react/src/utils/copyToClipboard.ts
```diff
@@ -0,0 +1,13 @@
+export function copyToClipboard(value: string) {
```
Hornebom, line 1: I you feel tempted, check for support of the `navigator.clipboard` in the client
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard

Hornebom (approved):

offerijns:
I was testing this locally.  I really like where this is going, very clear flow that covers a lot of use cases (Polkadot and EVM wallets, proxied and non proxied accounts, different EVM networks, ...).

A few smaller points I noticed:
- Just naming, I think more consistent with the rest of the app would be: `Choose Account` => `Choose account`
- I would expect to be able to click on the Network / Account icons when viewing the account list, to be able to switch them. Ideal behaviour in my mind is that if I click on the wallet icon, I go back to the previous screen and the network is still selected but the wallet is unselected, and when I click on the network icon, both are unselected and I basically start from scratch.
- For some proxy types we have rewrites to phrase them better, e.g. `Any` => `Any rights`. Can we also implement this for `PermissionManagement` => `Manage permissions`, and any other missing proxy types similarly? (I’m not sure which ones are already implemented) 
- In the account list, there is always a `|` between the account name (e.g. Alice) and the address. In the designs this dash is only present for proxied accounts, i.e. it’s a dash between the proxied account and the proxy account (aka the delegator and delegatee). I think this is important as it’s now hard to differentiate proxied and non proxied accounts.
- The dropdown below the connect button seems to be narrower, in the designs in Figma it is as wide as the Connect button.

onnovisser:
Based on the above feedback


> Just naming, I think more consistent with the rest of the app would be: `Choose Account` => `Choose account`

Copy changed

> I would expect to be able to click on the Network / Account icons when viewing the account list, to be able to switch them. Ideal behaviour in my mind is that if I click on the wallet icon, I go back to the previous screen and the network is still selected but the wallet is unselected, and when I click on the network icon, both are unselected and I basically start from scratch.

Yes I think I would expect the same as a user. Made the icons clickable and go back to either wallet or network selection

> For some proxy types we have rewrites to phrase them better, e.g. `Any` => `Any rights`. Can we also implement this for `PermissionManagement` => `Manage permissions`, and any other missing proxy types similarly? (I’m not sure which ones are already implemented)

Added copy for PermissionManagement and PodAuth types

> In the account list, there is always a `|` between the account name (e.g. Alice) and the address. In the designs this dash is only present for proxied accounts, i.e. it’s a dash between the proxied account and the proxy account (aka the delegator and delegatee). I think this is important as it’s now hard to differentiate proxied and non proxied accounts.

Removed from non-proxies

> The dropdown below the connect button seems to be narrower, in the designs in Figma it is as wide as the Connect button.

Made it a bit wider but it's a fixed width, so won't always match the width of the button

offerijns:
Awesome! Checked again, only thing I still noticed is that if I click on the wallet, then it goes back to theCchoose wallet dialog with the wallet unselected at the top, but Polkadot.js still selected at the bottom, see:

Screenshot 2023-02-20 at 18 46 35

Also, when I select Ttalisman which only has 1 connected account, the Switch wallet button in the dropdown is gone. I would expect it to still be shown in this case.

offerijns:
@onnovisser I tested again, everything above is now fixed! 💯

Hornebom (approved):

sophialittlejohn (approved):
:shipit:

PR #1183, opened 2023-02-07
Issuer details

Hornebom:
### Description
- extends create pool form with issuer details
- renders Issuer details in IssuerSection
- adds 'details' to pool metadata
- creates fabric Accordion and Accordion story
- extends Collapsible with general BoxProps

Ticket: https://app.zenhub.com/workspaces/centrifuge-applications-5fa29e924059e5001625e2a3/issues/gh/centrifuge/apps/1119

### Approvals

- [ ] Dev
- [ ] Product

### Test
- Create new pool and add (max 3) details right under the 'Custom content' title
- new created pool should show an accordion of these details

### Screenshots
create-details
issuer-details


annamehr:
Considerations:
1. What would be a descriptive title for this section on the 'create pool' page?
2. as I need both title and body for each title, I made both required > that might be an issue as I decided to display an empty detail by default to make it visible for the user. If the user don't want to use details, they would get an required error
3. In code this new metadata entry is named details > I'm not too sold on that and are happy for better options

1.descriptive title: Issuer profile 
2. I think it makes sense to have title and details as required. Maybe I don’t understand exactly what the problem is if we have both as required?
3. what about Issuer_details ?

onnovisser (approved):

PR #1192, opened 2023-02-09
OnboardingAPI: Add investor whitelisting

sophialittlejohn:
### Description

This pull request...

Closes #<issue-number>

### Approvals

- [x] Dev


centrifuge-app/src/utils/useProposalEstimate.ts
```diff
@@ -87,6 +87,7 @@ export function useProposalEstimate(formValues: Pick<PoolMetadataInput, 'tranche
         currency,
         CurrencyBalance.fromFloat(values.maxReserve || 0, chainDecimals),
         {} as any,
+        '4e',
```
jpangelle, line 90: is this a hex value for something?

onboarding-api/src/controllers/user/updateInvestorStatus.ts
```diff
@@ -28,6 +29,21 @@ export const updateInvestorStatusController = async (
...
+      return !step?.completed
+    })
+    if (incompleteSteps.length > 0) {
+      if (incompleteSteps) {
+        throw new HttpsError(
+          400,
+          `Incomplete onboarding steps for investor: ${incompleteSteps.map((step) => step[0]).join(', ')}`
+        )
+      }
+    }
```
jpangelle, line 45: You will have to add logic to check if the investor is US based or not and conditionally ignore the status `verifyAccreditation`. But perhaps this check isn't even necessary since the below block is ensuring that the `onboardingStatus` is `'pending'` which is only changed based on checks in the previous endpoint `/signAndSendDocuments`.

onboarding-api/src/controllers/user/updateInvestorStatus.ts
```diff
@@ -47,10 +63,13 @@ export const updateInvestorStatusController = async (
 
     if (user?.email && status === 'approved') {
       await sendApproveInvestorMessage(user.email, poolId, trancheId)
+      const hash = await whitelistInvestor(walletAddress, poolId, trancheId)
+      return res.status(200).send({ hash })
```
jpangelle, line 67: If you are changing the status code of the response, you will also need to change here: https://github.com/centrifuge/apps/blob/main/centrifuge-app/src/pages/Onboarding/UpdateInvestorStatus.tsx#L34

centrifuge-app/src/utils/useProposalEstimate.ts
sophialittlejohn: it's really just supposed to be a placeholder here, this hook is just used to get the paymentInfo (cost of tx) of a pool proposal (democracy). Your comment made me realize that this entire feature is broken because the democracy interface appears to have change on chain. I'll open a new PR after this to fix this feature!

onboarding-api/src/controllers/user/updateInvestorStatus.ts
sophialittlejohn: good catch! I'll include the conditional. The reason I added this check is to be extra sure that every single step is complete in case someone tried to hit the endpoint directly. Was thinking of it as kind of a fail safe, also for us in case we missed any other checks along the way

centrifuge-js/src/modules/pools.ts
```diff
@@ -629,6 +630,15 @@ export function getPoolsModule(inst: Centrifuge) {
                     PoolRole: 'LoanAdmin',
                   }
                 ),
+                // give the pure proxy permissions to whitelist investor for onboarding
```
offerijns, line 633: I don't think we should do this automatically on pool creation. Some pools might not want to use the automated onboarding. Rather we should build a UI feature to set this up. For the first pool(s) they could just manually add the account.

onboarding-api/src/utils/centrifuge.ts
```diff
@@ -1,5 +1,50 @@
...
+  const signer = keyring.addFromUri('//Dave')
+  const api = await ApiPromise.create({ provider: new WsProvider(process.env.COLLATOR_WSS_URL) })
+  const submittable = api.tx.permissions.add(
+    { PoolRole: 'MemberListAdmin' },
+    walletAddress,
+    { Pool: poolId },
+    { PoolRole: { TrancheInvestor: [trancheId, TenYearsFromNow] } }
+  )
+
+  // this doens't work b/c .connect() exepcts a Signer, not a KeyringPair
```
offerijns, line 37: Would be nice if we could adopt centrifuge.js to support this.

onboarding-api/src/utils/centrifuge.ts
```diff
@@ -1,5 +1,50 @@
 import Centrifuge from '@centrifuge/centrifuge-js'
+import { ApiPromise, WsProvider } from '@polkadot/api'
+import { Keyring } from '@polkadot/keyring'
+import { cryptoWaitReady } from '@polkadot/util-crypto'
+import { firstValueFrom } from 'rxjs'
+
+const TenYearsFromNow = Math.floor(Date.now() / 1000 + 10 * 365 * 24 * 60 * 60)
```
offerijns, line 7: I would even make this 100 years from now :) At least that's what we did in Tinlake. Or even the max uint64 value.

onboarding-api/src/utils/centrifuge.ts
```diff
@@ -1,5 +1,50 @@
...
 })
+
+export const getPoolById = async (poolId: string) => {
+  const pools = await firstValueFrom(centrifuge.pools.getPools())
+  const pool = pools.find((p) => p.id === poolId)
+  const metadata = await firstValueFrom(centrifuge.metadata.getMetadata(pool?.metadata!))
+  return { pool, metadata }
+}
+
+export const whitelistInvestor = async (walletAddress: string, poolId: string, trancheId: string) => {
```
offerijns, line 24:
```suggestion
export const addInvestorToMemberList = async (walletAddress: string, poolId: string, trancheId: string) => {
```

offerijns (commented):
Logic of adding investors looks good to me!

onboarding-api/src/utils/centrifuge.ts
sophialittlejohn: for sure won't expire then 😄  will also change this in the UI where investors are given 10 years atm

onboarding-api/src/utils/centrifuge.ts
sophialittlejohn: Agreed! Seems like it will require a bigger change on cent-js (potentially affecting the entire app) so I'll tackle it in another PR

onboarding-api/src/controllers/user/updateInvestorStatus.ts
```diff
@@ -33,6 +33,9 @@ export const updateInvestorStatusController = async (
       if (name === 'signAgreements') {
         return !step?.[poolId]?.[trancheId]?.signedDocument
       }
+      if (!user.countryOfCitizenship?.startsWith('us') && name === 'verifyAccreditation') {
+        return true
+      }
```
jpangelle, line 38: Unfortunately, it is a little more complicated based on the investor type:

```ts
if (
  name === "verifyAccreditation" &&
  ((user.investorType === "individual" && user.countryOfCitizenship !== "us") ||
    (user.investorType === "entity " &&
      !user.jurisdictionCode.startsWith("us")))
) {
  return true;
}
```

onboarding-api/src/controllers/user/updateInvestorStatus.ts
```diff
@@ -46,11 +65,14 @@ export const updateInvestorStatusController = async (
     await validateAndWriteToFirestore(walletAddress, updatedUser, 'entity', ['onboardingStatus'])
 
     if (user?.email && status === 'approved') {
+      await addInvestorToMemberList(walletAddress, poolId, trancheId)
       await sendApproveInvestorMessage(user.email, poolId, trancheId)
+      return res.status(204).send()
```
jpangelle, line 70: 👍

onboarding-api/src/controllers/user/updateInvestorStatus.ts
sophialittlejohn: oops, thanks!

jpangelle (approved):
🍭 🍭 🍭

PR #1193, opened 2023-02-10
clean up signing logic

jpangelle:
### Description

This pull request does the following:
- renames `useOnboardingUser` to `useOnboarding` so that the hook can also include the information of the pool that they are onboarding to
- renames `OnboardingUserProvider` to `OnboardingProvider`
- removes all hardcoded instances of pool information and move to `OnboardingProvider` context
- refactors and simplifies the logic of how the signing flow and emailing work
- updates types to include `transactionInfo` which is an object composed of `extrinsicHash` and `blockNumber`
- verifies the that the remark signed on chain is valid
- adds `null`s to type definition

### Approvals

- [ ] Dev


onboarding-api/src/controllers/agreement/signAgreement.ts
```diff
@@ -46,8 +46,7 @@ export const signAgreementController = async (
     const pages = pdfDoc.getPages()
     const lastPage = pages[pages.length - 1]
 
-    // @ts-expect-error
-    lastPage.drawText(user?.name, {
+    lastPage.drawText(user?.name as string, {
```
jpangelle, line 49: I think casting to `string` better conveys the idea that we are confident that `user.name` will be available.

centrifuge-app/src/pages/Onboarding/BusinessOwnership.tsx
```diff
@@ -137,11 +119,33 @@ const BusinessOwnershipInlineFeedback = ({ isError }: { isError: boolean }) => {
 
 export const BusinessOwnership = ({ backStep, nextStep }: Props) => {
   const { authToken } = useAuth()
-  const { onboardingUser, refetchOnboardingUser } = useOnboardingUser() as {
+  const { onboardingUser, refetchOnboardingUser, pool } = useOnboarding() as {
```
sophialittlejohn, line 122: could `refetchOnFocus` be an option you pass to the hook? then you can allow the `useQuery` hook to handle the refetching automatically instead of attaching your own event listeners

onboarding-api/src/utils/validateRemark.ts
```diff
@@ -0,0 +1,20 @@
+import { firstValueFrom } from 'rxjs'
+import { InferType } from 'yup'
+import { sendDocumentsToIssuerInput } from '../controllers/emails/sendDocumentsToIssuer'
+import { centrifuge } from './centrifuge'
+import { HttpsError } from './httpsError'
+
+export const validateRemark = async (
```
sophialittlejohn, line 7: nice!!

centrifuge-app/src/pages/Onboarding/BusinessOwnership.tsx
jpangelle: So, `useQuery` has the `refetchOnWindowFocus` option but since the implementation of this `useQuery` is located in Context, there is only one single instance. If I were to pass in arguments to change the behavior of the `refetchOnWindowFocus`, it would apply to all the other component as well.

sophialittlejohn: makes sense, makes me wonder if context is actually necessary. Wondering if it might be easier and more flexible if it were a simple hook with state instead

jpangelle: will revisit when cleaning up

sophialittlejohn (approved):

PR #1196, opened 2023-02-13
migrate react-query to folder

jpangelle:
### Description

This pull request:
- combines the `useOnboarding` and the `useOnboardingStep` hook
- migrates the `useMutation`s and `useQuery`s to a `queries` folder
- standardizes query keys

### Approvals

- [x] Dev


centrifuge-app/src/components/OnboardingProvider.tsx
```diff
@@ -16,11 +17,21 @@ const OnboardingContext = React.createContext<{
...
+  nextStep: () => void
+  previousStep: () => void
+  isFetchingStep: boolean
 } | null>(null)
 
 export function OnboardingProvider({ children }: { children?: React.ReactNode }) {
   const { authToken } = useAuth(AUTHORIZED_ONBOARDING_PROXY_TYPES)
-  const { selectedAccount } = useWallet()
+  const { isConnecting, selectedAccount } = useWallet()
+  const { isAuth, isAuthFetched } = useAuth(AUTHORIZED_ONBOARDING_PROXY_TYPES)
```
sophialittlejohn, line 30: I see `useAuth` twice here!

centrifuge-app/src/pages/Onboarding/BusinessInformation.tsx
```diff
@@ -56,12 +49,11 @@ const BusinessInformationInlineFeedback = ({ isError }: { isError: boolean }) =>
   return null
 }
 
-export const BusinessInformation = ({ backStep, nextStep }: Props) => {
-  const { authToken } = useAuth()
-  const { onboardingUser, refetchOnboardingUser, pool } = useOnboarding() as {
+export const BusinessInformation = () => {
+  const { onboardingUser, previousStep, nextStep } = useOnboarding() as {
```
sophialittlejohn, line 53: maybe you could add types to the hook instead of casting? 🙃

sophialittlejohn (commented):
In trying to find the balance between DRYness and ease of reading, I feel like the hooks that fetch from the onboarding api could be combined into one hook that receives options whose only purpose is to send and receive data from the API (`useOnboardingApi()` for example). Then the logic associated with specific steps in the FE (e.g signRemark) could live local to that step 🤔  To me, a hook for each individual step seems a little much because they will not be reused ever

centrifuge-app/src/components/OnboardingProvider.tsx
jpangelle: wow 🤦

centrifuge-app/src/pages/Onboarding/BusinessInformation.tsx
jpangelle: Yeah this one is tough because the type for `onboardingUser` is `IndividualUser | EntityUser` but if the UI is rendering this component then we can be sure that the type is `EntityUser` however TS is not smart enough to infer this. Perhaps I need to rethink it to see if I can refactor it in a way where TS can infer.

jpangelle:
> In trying to find the balance between DRYness and ease of reading, I feel like the hooks that fetch from the onboarding api could be combined into one hook that receives options whose only purpose is to send and receive data from the API (`useOnboardingApi()` for example). Then the logic associated with specific steps in the FE (e.g signRemark) could live local to that step 🤔 To me, a hook for each individual step seems a little much because they will not be reused ever

Love it! 😍 Also, how did that not occur to me as I was basically making a bunch of duplicate hooks lol?!

centrifuge-app/src/pages/Onboarding/BusinessInformation.tsx
jpangelle: This [commit](https://github.com/centrifuge/apps/pull/1196/commits/ceb4903d47e5d391af29cb7dea6dabc319bddbd5) still uses type casting but in a way where only the `onboardingUser` needs to be type casted. I tried a few different approaches and couldn't come up with anything cleaner than this (having two different onboarding hooks, having the hook accept an argument of `investorType`, adding all of the `entity` properties to the `individual` type as `undefined`, etc.)

sophialittlejohn: did you try allowing your hook to accept a generic type? when you know for sure its the `EntityUser` you could pass that as the generic type. But, I thought it should be possible for ts to infer the type of user based on its discriminated union (`investorType`). The only time the discriminated union is difficult is when you want to destructure properties before inferring the type

jpangelle: The discriminated union logic would only apply if I wrapped the usage in an `if` statement:

```tsx
export const BusinessInformation = () => {
  const { onboardingUser } = useOnboarding()
  if (onboardingUser.investorType === 'entity') {
    const formik = useFormik({
      initialValues: {
        businessName: onboardingUser.businessName || '',
        // other initial values
      }
    })
  
    // other component logic
  }
}
```
But with the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level), you can't wrap hooks inside conditionals so the `useFormik` would break that rule.

I tried my hand at using a generic and optionally passing in a type but I could not figure it out or if it was even possible. Perhaps you can give it a shot 🙏

sophialittlejohn (approved):
💯

PR #1205, opened 2023-02-17
CentrifugeApp: Fix pool proposal and preimage noting

sophialittlejohn:
### TODO
- fix proposal fee estimate (find const for for preimage deposit amount)

### Approvals

- [ ] Dev


jpangelle (approved):
A bit over my head but looks good from a high level 😸

centrifuge-app/src/utils/useProposalEstimate.ts
```diff
@@ -31,19 +70,22 @@ export function useProposalEstimate(formValues: Pick<PoolMetadataInput, 'tranche
...
+              const preimageFee = preimageByteDeposit
+                // the first argument passed to the `notePreimage` extrinsic is the actual encoded proposal in bytes
+                .mul(new BN((submittable as any).method.args[0].length))
+                .add(preimageBaseDeposit)
+
+              return new CurrencyBalance(preimageFee, chainDecimals)
+            } else if (config.poolCreationType === 'propose') {
+              return new CurrencyBalance(hexToBN(minimumDeposit.toHex()), chainDecimals)
+            }
+            return new CurrencyBalance(0, chainDecimals)
```
onnovisser, line 88: not sure why we're setting the chain decimals in state, we're only using it here

centrifuge-app/src/utils/useProposalEstimate.ts
```diff
@@ -76,17 +118,15 @@ export function useProposalEstimate(formValues: Pick<PoolMetadataInput, 'tranche
...
       feeSubject.next([
         selectedAccount.address,
         '1234567890',
         '1234567890',
         tranches,
-        currency,
+        'ausd',
         CurrencyBalance.fromFloat(values.maxReserve || 0, chainDecimals),
-        {} as any,
+        mockMetadata as any,
```
onnovisser, line 129: to get a fee, we only need the tranches and poolName of the metadata, but we should get the based on the form values because it needs the same number of tranches as the user has enabled in the form. 
something like:
```
{ poolName: values.poolName ?? 'Poolio',  tranches: values.tranches.map(t => ({
    tokenName: t.tokenName || 'J',
    symbolName: t.symbolName || 'JJJ'
  }) 
} as any
```
these values also influence the fee. The longer the values, the higher the fee

PR #1206, opened 2023-02-17
onboarding improvements

jpangelle:
### Description

This pull request:
  - adds a `Country of Residence` field to KYC step
  - conditionally renders text for KYC screen based on investor type
  - refactors the tax `POST` and `GET` endpoints to not nest the documents in `pool` and `tranche` ids
  - makes schemas more accurate

### Approvals

- [x] Dev


centrifuge-app/src/pages/Onboarding/KnowYourCustomer/SignerVerification.tsx
```diff
@@ -21,16 +22,30 @@ const formatCountryCodes = (countryCodes: { [key: string]: string }) => {
...
+    title: 'Signer verification',
+    description: 'Please add your information to complete verification.',
+    checkboxLabel: 'I confirm that all the information provided is true and accurate.',
+  },
+}
+
+export const SignerVerification = ({ formik, isLoading, isCompleted }: Props) => {
+  const { previousStep, nextStep, onboardingUser } = useOnboarding()
+
+  const investorType = onboardingUser?.investorType === 'entity' ? 'entity' : 'individual'
```
jpangelle, line 42: At this point, `onboardingUser.investor` is either `entity` or `undefined`.

centrifuge-app/src/pages/Onboarding/KnowYourCustomer/index.tsx
```diff
@@ -4,13 +4,14 @@ import { boolean, date, object, string } from 'yup'
 import { useOnboarding } from '../../../components/OnboardingProvider'
 import { useStartKYC } from '../queries/useStartKYC'
 import { useVerifyIdentity } from '../queries/useVerifyIdentity'
-import { AuthorizedSignerVerification } from './AuthorizedSignerVerification'
 import { IdentityVerification } from './IdentityVerification'
+import { SignerVerification } from './SignerVerification'
 
-const authorizedSignerInput = object({
+const signerInput = object({
```
jpangelle, line 10: Renamed all uses of `authorizedSigner` to just `signer` so that it makes sense for either `entity` OR `individual`

centrifuge-app/src/pages/Onboarding/queries/useTaxInfo.ts
```diff
@@ -6,22 +6,19 @@ import { useOnboarding } from '../../../components/OnboardingProvider'
...
-        }
-      )
+      const response = await fetch(`${import.meta.env.REACT_APP_ONBOARDING_API_URL}/getTaxInfo`, {
+        method: 'GET',
+        headers: {
+          Authorization: `Bearer ${authToken}`,
+          'Content-Type': 'application/json',
+        },
+        credentials: 'include',
+      })
```
jpangelle, line 21: No longer sends `poolId` or `trancheId` as parameters since we will not longer nest the document in folders and instead it will just be `${walletAddress}.pdf`

centrifuge-app/src/pages/Onboarding/queries/useVerifyEmail.ts
```diff
@@ -1,16 +1,14 @@
...
 import { useAuth } from '../../../components/AuthProvider'
 
 export const useVerifyEmail = () => {
   const { authToken } = useAuth()
   const { search } = useLocation()
-  const { selectedAccount } = useWallet()
   const token = new URLSearchParams(search).get('token')
 
   const query = useQuery(
-    ['verify-email', selectedAccount?.address],
```
jpangelle, line 13: Was causing unnecessary refetches. My baddd!

centrifuge-app/src/types/index.ts
```diff
@@ -86,7 +86,7 @@ type IndividualUserSteps = {
 export interface EntityOnboardingSteps extends IndividualUserSteps {
   verifyBusiness: {
     completed: boolean
-    timeStamp: string | null
+    timeStamp: string
```
jpangelle, line 89: This value will always be defined since `verifyBusiness` is added in the first endpoint hit in the flow.

centrifuge-app/src/types/index.ts
```diff
@@ -102,7 +102,7 @@ export type EntityUser = {
   investorType: 'entity'
   walletAddress: string
   businessName: string
-  email: string | null
+  email: string
```
jpangelle, line 105: Same here, `email` will always be a `string`.

centrifuge-app/src/types/index.ts
```diff
@@ -124,9 +125,10 @@ export type EntityUser = {
 type IndividualUser = {
   investorType: 'individual'
   walletAddress: string
-  name: string | null
-  dateOfBirth: string | null
-  countryOfCitizenship: string | null
+  name: string
+  dateOfBirth: string
+  countryOfCitizenship: string
+  countryOfResidence: string
```
jpangelle, line 131: For the `individual` flow, the user is created with these values so they will never be `null`.

onboarding-api/src/database/index.ts
```diff
@@ -77,7 +77,7 @@ export const entityUserSchema = object({
   investorType: string().default('entity') as StringSchema<Entity>,
   wallet: walletSchema,
   kycReference: string().optional(),
-  email: string().email().default(null),
+  email: string().email().required(),
```
jpangelle, line 80: In `entity` flow, `email` will never be `null`.

onboarding-api/src/database/index.ts
```diff
@@ -113,10 +114,11 @@ export const individualUserSchema = object({
...
   kycReference: string().optional(),
-  email: string().default(null).nullable(),
-  name: string().nullable().default(null),
-  dateOfBirth: string().nullable().default(null),
-  countryOfCitizenship: string().nullable().default(null), // TODO: validate with list of countries
+  email: string().default(null).nullable(), // TODO: coming soon
+  name: string().required(),
+  dateOfBirth: string().required(),
+  countryOfCitizenship: string().required(), // TODO: validate with list of countries
+  countryOfResidence: string().required(), // TODO: validate with list of countries
```
jpangelle, line 121: In `individual` flow, these values will never be `null`

centrifuge-react/src/components/WalletProvider/WalletProvider.tsx
```diff
@@ -54,7 +54,7 @@ export function WalletProvider({ children }: WalletProviderProps) {
   const [accounts, setAccounts] = React.useState<Account[] | null>(null)
   const [selectedAccountAddress, setSelectedAccountAddress] = React.useState<string | null>(null)
   const [proxyAddress, setProxyAddress] = React.useState<string | null>(null)
-  const [isConnecting, setIsConnecting] = React.useState(false)
+  const [isConnecting, setIsConnecting] = React.useState(!triedEager && !!localStorage.getItem('web3Persist'))
```
jpangelle, line 57: @onnovisser Can you double check the changes in this file plz

onnovisser: This file was changed drastically for the EVM wallet integration https://github.com/centrifuge/apps/pull/1182/files#diff-7ce668a73451bfac7a8c6285f2ffcd3cf52b26bbd3128f49d44d09d000e2e4fe
so this will only cause conflicts. we can find a way to reimplement these changes in that PR instead of here

onnovisser: for example the new hook `useConnectEagerly` could return a boolean if it's trying to connect eagerly and we can combine that value with isConnecting
https://github.com/centrifuge/apps/pull/1182/files#diff-7ce668a73451bfac7a8c6285f2ffcd3cf52b26bbd3128f49d44d09d000e2e4fe

jpangelle: 👍 removed the code from this PR

centrifuge-app/src/pages/Onboarding/KnowYourCustomer/index.tsx
sophialittlejohn: hmm 🤔  maybe KYC would make more sense for naming. Signer could be kind of confusing since we use the concept of signing for transactions too. And in the individual use case it would be called KYC, right?

onboarding-api/src/database/index.ts
```diff
@@ -172,7 +174,7 @@ export const validateAndWriteToFirestore = async <T = undefined | string[]>(
   key: string,
   data: T extends 'undefined' ? OnboardingUser : Subset<OnboardingUser>,
   schemaKey: keyof typeof schemas,
-  mergeFields?: T
+  mergeFields?: string[]
```
sophialittlejohn, line 177: why this change? the generic type `T` should be used to determine the type of data being passed: whether a `Subset` is allowed of the entire structure must be passed

centrifuge-app/src/pages/Onboarding/KnowYourCustomer/index.tsx
jpangelle: The KYC step is really two substeps.

1. Signer Verification
2. Identity Verification (ShuftiPro)

and both of those components live in the `KnowYourCustomer` folder.

Maybe I can call it `WalletSignerVerification`?

onboarding-api/src/database/index.ts
jpangelle: I was trying to resolve this Typescript error without `@ts-expect-error` and this resolved the issue. I thought it made sense because `T` is `undefined | string[]` which is the same as `[property]?: string[]` or so I thought.

image

sophialittlejohn: weird, I'm not getting that error in my editor. Maybe we're using different versions of typescript. But also, it doesn't look it works anymore anyway 😅

centrifuge-app/src/pages/Onboarding/KnowYourCustomer/index.tsx
sophialittlejohn: Maybe I misunderstood then, if this is actually the verification for the signer (the wallet signing the tx) then signer does make sense!

jpangelle: yeah exactly!

onboarding-api/src/database/index.ts
jpangelle: reverted

sophialittlejohn (approved):

PR #1216, opened 2023-02-22
Patch GHA deploy to demo logic

gpmayorga:
Quick fix on the logic that would make demo deployments point to DEV instead (demo is not being updated).


sophialittlejohn (approved):
🎉

PR #1217, opened 2023-02-22
fix: redirect urls

jpangelle:
### Description

This pull request fixes the redirect urls. Without the `scheme`, the verify email button in the verification email does not work. There was also an incorrect trailing `s` on `app`.

### Approvals

- [ ] Dev


sophialittlejohn (approved):
oops, thanks for fixing 🙌

PR #1232, opened 2023-03-07
enable click on chevron

Hornebom:
### Approvals
- [ ] Dev


sophialittlejohn (approved):
Definitely better than before 🙌

PR #1236, opened 2023-03-08
rewording Epoch to Order

Hornebom:
### Description
Ticket https://app.zenhub.com/workspaces/centrifuge-applications-5fa29e924059e5001625e2a3/issues/gh/centrifuge/apps/1201

### Approvals
- [ ] Dev
- [ ] Product


centrifuge-app/src/components/LiquidityTransactionsSection.tsx
```diff
@@ -116,7 +116,7 @@ export function LiquidityTransactionsSection({
           data={chartData}
           names={dataNames}
           colors={dataColors}
-          xAxisLabel="Latest epoch"
+          xAxisLabel="Latest order"
```
offerijns, line 119: Currently this chart is still grouped by epoch. We're not renaming epochs to orders, rather epochs are "sets / groups of orders" and we're trying to change terminology to make it work this way.

For this chart, I think what would be best is rather than grouping it by epoch, grouping it by day, and showing `Latest day` and the date on the order days on the axis. I think the subquery should also have the equivalent data per day already.

centrifuge-app/src/components/LiquidityEpochSection.tsx
```diff
@@ -43,11 +44,11 @@ const EpochStatusOngoing: React.FC<{ pool: Pool }> = ({ pool }) => {
     useLiquidity(pool.id)
   const { message: epochTimeRemaining } = useEpochTimeCountdown(pool.id)
   const { execute: closeEpochTx, isLoading: loadingClose } = useCentrifugeTransaction(
-    'Close epoch',
+    'Close order',
```
offerijns, line 47: This is also technically incorrect as a user is not `closing an order`. Maybe we change this to `Start order execution` or something like that, wdyt @annamehr ?

centrifuge-app/src/components/LiquidityEpochSection.tsx
```diff
@@ -43,11 +44,11 @@ const EpochStatusOngoing: React.FC<{ pool: Pool }> = ({ pool }) => {
...
     (cent) => cent.pools.closeEpoch,
     {
       onSuccess: () => {
-        console.log('Epoch closed successfully')
+        console.log('Order closed successfully')
```
offerijns, line 51: Likewise

centrifuge-app/src/components/LiquidityEpochSection.tsx
```diff
@@ -70,22 +71,43 @@ const EpochStatusOngoing: React.FC<{ pool: Pool }> = ({ pool }) => {
 
   return (
     <PageSection
-      title={`Epoch ${pool.epoch.current}`}
+      title="Order overview"
       titleAddition={<Text variant="body2">{epochTimeRemaining ? 'Ongoing' : 'Minimum duration ended'}</Text>}
```
offerijns, line 75: With the removal of epochs, I think the `Ongoing` and `Minimum duration ended` subtitles have also become confusing. I would just remove this, if you look at the screenshot, the line below it already explains what's happening ("The epoch is continuing until orders have been locked and can be executed.").

Screenshot 2023-03-09 at 09 48 33

centrifuge-app/src/components/LiquidityEpochSection.tsx
annamehr: `Started order execution successfully`

offerijns: Perfect 👍

onnovisser (approved):

annamehr:
@Hornebom in the current implementation it says `Execution order` instead of `Start order execution`. can you change it to the latter? 
Screenshot 2023-03-16 at 14 06 53

PR #1239, opened 2023-03-10
Onboarding: Integrate EVM authentication

sophialittlejohn:
### Description

This pull request...
- allows onboarding users to onboard with their EVM wallet or substrate wallet 
- changes the authentication to validate either a `siwe` message for EVM or a jw3t token for substrate and produces a jwt token in response to use as API authentication
- removes the `AuthProvider` in favor of  `OnboardingAuthProvider` so the FE can authenticate for either wallet
- disconnects wallets when switching account
- adds support for multiple wallets in the API
- provides an API endpoint to verify if a JWT token is valid

Closes #<issue-number>

### Approvals

- [x] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product


onboarding-api/src/utils/fetchUser.ts
```diff
@@ -8,14 +8,14 @@ type UserOrNull<T> = T extends Options ? OnboardingUser | null : OnboardingUser
 
 export async function fetchUser<T>(walletAddress: string, options?: OptionsOrNever<T>): Promise<UserOrNull<T>> {
   try {
-    const userDoc = await userCollection.doc(walletAddress).get()
-    if (!userDoc.exists) {
+    const userSnapshot = await userCollection.where('wallet.address', '==', walletAddress).get()
```
sophialittlejohn, line 11: this change will allow us to query the user by connected wallet, so they could have multiple wallets associated with one onboarding user. Supposedly the query shouldn't take longer than querying by unique ID 🙌

onboarding-api/src/utils/shuftiProRequest.ts
```diff
@@ -5,7 +5,10 @@ import { kybMockResponse } from '../mocks/kybResponse'
 import { HttpError } from './HttpError'
 
 export const shuftiProRequest = async (_req: Request, payload: any, options?: { dryRun?: boolean; path?: string }) => {
-  if (options?.dryRun) {
+  if (
+    options?.dryRun &&
+    (process.env.COLLATOR_WSS_URL.includes('demo') || process.env.COLLATOR_WSS_URL.includes('development'))
+  ) {
```
sophialittlejohn, line 11: Added some protection for the dryRun parameter

centrifuge-app/src/pages/Onboarding/LinkWallet.tsx
```diff
@@ -18,9 +16,11 @@ const validationSchema = object({
 export const LinkWallet = () => {
   const [isDataSharingAgreementDialogOpen, setIsDataSharingAgreementDialogOpen] = React.useState(false)
   const { nextStep } = useOnboarding()
-
-  const { selectedAccount } = useWallet().substrate
-  const { login, isAuth } = useAuth()
+  const { login, isAuth } = useOnboardingAuth()
```
onnovisser, line 19: there's more places that used `useAuth` in the onboarding parts. do they also need updating? Do we still need the AuthProvider? As far as I know it's only used in onboarding

sophialittlejohn: Thanks, updating the rest. I think the `useAuth` hook is also used for pod authentication

onnovisser: the POD stuff uses its own PodAuthProvider

onboarding-api/src/controllers/auth/refreshToken.ts
```diff
@@ -0,0 +1,28 @@
+import { Request, Response } from 'express'
+import * as jwt from 'jsonwebtoken'
+import { SupportedNetworks } from '../../database'
+import { HttpError, reportHttpError } from '../../utils/httpError'
+
+export const refreshTokenController = async (req: Request, res: Response) => {
```
onnovisser, line 6: jwt's get some security from the fact that they expire. Infinitely refreshing tokens without using a refresh token that can be revoked doesn't seem very secure

centrifuge-app/src/pages/Onboarding/LinkWallet.tsx
sophialittlejohn: two hook with the same name 😱  confusing!

onboarding-api/src/controllers/auth/refreshToken.ts
sophialittlejohn: so this is not infinite because the jw3t token from the signed message expires which would make the refresh endpoint fail

onnovisser: but this is verifying the jwt token, not the jw3t one

onboarding-api/src/controllers/auth/generateNonce.ts
```diff
@@ -0,0 +1,6 @@
+import { Request, Response } from 'express'
+import { generateNonce } from 'siwe'
+
+export const generateNonceController = (req: Request, res: Response) => {
+  return res.status(200).send({ nonce: generateNonce() })
```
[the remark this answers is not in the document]
onnovisser: I meant, for the nonce to be useful, it needs to be stored in the user session (like a cookie) and checked against the nonce provided in the signed message.

onboarding-api/src/controllers/auth/refreshToken.ts
sophialittlejohn: you right! Seems better to remove the refresh functionality and make the user resign if the jwt expires

onboarding-api/src/controllers/auth/generateNonce.ts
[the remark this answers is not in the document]
sophialittlejohn: @onnovisser thanks! I took your advice and changed the implementation. For eth wallets now the nonce and wallet address are sent in an encrypted httpOnly cookie. Before sending the auth token to the client we check to make sure the wallet and nonce match the ones used for when generating the nonce and signing the message. Let me know what you think of this solution

centrifuge-app/vite.config.ts
```diff
@@ -21,7 +21,6 @@ export default defineConfig({
       // ensure all packages are converted to ES6 for rollup bundle
       include: [/node_modules/],
     },
-    sourcemap: true,
```
sophialittlejohn, line 24: this only disables source maps in production

centrifuge-app/src/components/OnboardingProvider.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react'
 import { useQuery } from 'react-query'
 import { OnboardingUser } from '../types'
 import { getActiveOnboardingStep } from '../utils/getActiveOnboardingStep'
-import { useAuth } from './AuthProvider'
+import { useOnboardingAuth } from './OnboardingAuthProvider'
 
 const AUTHORIZED_ONBOARDING_PROXY_TYPES = ['Any', 'Invest', 'NonTransfer', 'NonProxy']
```
jpangelle, line 8: I guess you can remove 🤷

onboarding-api/src/controllers/auth/authenticateWallet.ts
```diff
@@ -0,0 +1,85 @@
...
+  jw3tToken: string(),
+})
+
+export const authenticateWalletController = async (
+  req: Request<{}, {}, InferType<typeof verifyWalletInput>>,
+  res: Response
+) => {
+  try {
+    await validateInput(req.body, verifyWalletInput)
+    const payload = req.body.jw3tToken ? await verifySubstrateWallet(req) : await verifyEthWallet(req, res)
```
jpangelle, line 29: `jw3tToken` aka `JSON Web3 Token Token` 😅

onboarding-api/src/controllers/auth/authenticateWallet.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+  return {
+    address,
+    network: 'substrate',
+  }
+}
+
+async function verifyEthWallet(req: Request, res: Response) {
+  try {
+    const key = req.signedCookies['onboarding-auth']
```
jpangelle, line 66: Is the name of the cookie is `onboarding-auth`? If so, should it have a more unique name like maybe appending the wallet address in the same way that we do for the j3wt token?

onboarding-api/src/utils/fetchUser.ts
```diff
@@ -6,16 +7,16 @@ type Options = { suppressError?: boolean }
...
 type UserOrNull<T> = T extends Options ? OnboardingUser | null : OnboardingUser
 
-export async function fetchUser<T>(walletAddress: string, options?: OptionsOrNever<T>): Promise<UserOrNull<T>> {
+export async function fetchUser<T>(wallet: Request['wallet'], options?: OptionsOrNever<T>): Promise<UserOrNull<T>> {
   try {
-    const userDoc = await userCollection.doc(walletAddress).get()
-    if (!userDoc.exists) {
+    const userSnapshot = await userCollection.where(`wallet`, 'array-contains', wallet).get()
+    if (userSnapshot.empty) {
       if (options && options.suppressError) {
```
jpangelle, line 14: I think `options && options.suppressError` is the same as `options?.suppressError`

jpangelle (approved):
just a few minor questions and comments.

INCREDIBLE JOB!! 😎

onboarding-api/src/controllers/auth/authenticateWallet.ts
sophialittlejohn: good point, just in case I appended the address to the cookie name instead of attaching it to the cookie value

PR #1245, opened 2023-03-15
Feat/wallet component feedback 1

Hornebom:
### Description
From https://app.zenhub.com/workspaces/centrifuge-applications-5fa29e924059e5001625e2a3/issues/gh/centrifuge/apps/1167 this pr solves the following:
- display the current pool issuer name in the invest cta box
- implements smaller design changes to the wallet dropdown (view on subscan not available in dev environment)

Not implemented:
- display values in USD
- display the wallet button above its dropdown menu

### Approvals

- [ ] Dev
- [ ] Product


annamehr:
what happened to this bug? should I move it to a new bug ticket?

> Another small bug when I haven't invested in the pool yet and the token dropdown menu displays twice the same token instead of displaying the junior and senior token.

image

centrifuge-react/src/components/WalletMenu/WalletMenu.tsx
```diff
@@ -58,6 +59,8 @@ function ConnectedMenu() {
       : balances
       ? [balances.native.balance, balances.native.currency.symbol]
       : []
+  const explorer = connectedNetwork && useGetExplorerUrl(connectedNetwork)
+  const subScanUrl = explorer && explorer.address(address)
```
onnovisser, line 63: can't conditionally call a hook. there will always be a connectedNetwork here, but if you want to be safe you can do:
```
const explorer = useGetExplorerUrl()
const subScanUrl = explorer.address(address, connectedNetwork)
```

Hornebom: true 👌  updated

onnovisser (approved):

annamehr (approved):
LGTM! 🚀

PR #1256, opened 2023-03-20
Centrifuge App: Get proxies directly from chain

onnovisser:
### Description

Gets proxies directly from the chain instead of the SubQuery

### Approvals

- [ ] Dev

### DIFF
--- a/centrifuge-js/src/modules/proxies.ts
+++ b/centrifuge-js/src/modules/proxies.ts
@@ -5,6 +5,8 @@ import { Account } from '../types'
 import { addressToHex } from '../utils'
 
 export function getProxiesModule(inst: CentrifugeBase) {
+  // TODO: Probably remove this as getting all proxies from the chain could potentially be quite a lot of data
+  // Used as a fallback for when the SubQuery is down
   function getAllProxies() {
     const $api = inst.getApi()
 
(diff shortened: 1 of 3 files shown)


centrifuge-js/src/modules/proxies.ts
```diff
@@ -2,8 +2,45 @@ import { encodeAddress } from '@polkadot/util-crypto'
...
+          })
+          .forEach((proxy) => {
+            const index = proxiesByDelegate[proxy.delegatee]?.findIndex((p) => p.delegator === proxy.delegator)
+            if (index > -1) {
+              proxiesByDelegate[proxy.delegatee][index].types.push(proxy.proxyType)
+            } else {
+              ;(proxiesByDelegate[proxy.delegatee] || (proxiesByDelegate[proxy.delegatee] = [])).push({
+                delegator: proxy.delegator,
+                delegatee: proxy.delegatee,
+                types: [proxy.proxyType],
```
offerijns, line 35: Would this work if there are multiple proxies set? Seems like if either it should push to the `types` field, or this should not be an array.

offerijns (approved):

centrifuge-js/src/modules/proxies.ts
onnovisser: it pushes on line 30 if that is the case

PR #1274, opened 2023-03-27
comment out uses of siwe

jpangelle:
### Description

This pull request migrates siwe to ssx in order to fix a https://github.com/spruceid/siwe/issues/154

### Approvals

- [ ] Dev


onboarding-api/src/controllers/auth/generateNonce.ts
```diff
@@ -23,9 +23,9 @@ export const generateNonceController = async (
     await validateInput(req.body, generateNonceInput)
     const nonce = generateNonce()
     res.cookie(`onboarding-auth-${req.body.address.toLowerCase()}`, `${nonce}`, {
-      secure: !(process.env.COLLATOR_WSS_URL.includes('demo') || process.env.COLLATOR_WSS_URL.includes('dev')),
+      secure: true,
       httpOnly: true,
-      sameSite: 'strict',
+      sameSite: 'none',
```
jpangelle, line 28: Since our client and server are are different domains, this has to be set to `none`. If it is set to `none`, the `secure` property has to be set to `true.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#fixing_common_warnings

sophialittlejohn: nice catch! `secure: true` usually doesn't work for localhost since it's by default not a secure connection, I think actually a better check for the secure key would be
```js
  secure: process.env.REDIRECT_URL.includes('localhost')  ? false : true
```
because it actually depends in which env the cookie is being set

centrifuge-app/src/components/OnboardingAuthProvider.tsx
```diff
@@ -180,27 +179,29 @@ const loginWithEvm = async (address: string, signer: any) => {
...
-  const signedMessage = await signer?.signMessage(siweMessage)
+
+  const message = `${domain} wants you to sign in with your Ethereum account:
+${address}
+
+Please sign to authenticate your wallet
+
+URI: ${origin}
+Version: 1
+Chain ID: 1
```
sophialittlejohn, line 193: I just learned this recently: since in our dev env we're using the goerli testnet we should use the chain id 5 here. Wonder if we could check in the provider for the chain id and pass it programmatically?

jpangelle: addressed

sophialittlejohn (approved):
💯  nice work @jpangelle!

PR #1295, opened 2023-04-04
Centrifuge App: Fix loan rates

onnovisser:
### Description

All rates related to loan are now yearly and no longer as rate per second.

### Approvals

- [ ] Dev


centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
```diff
@@ -354,22 +358,13 @@ function IssuerCreateLoan() {
...
-                        value={field.value}
-                        onChange={(event) => {
-                          form.setFieldValue('templateId', event.target.value)
-                        }}
-                        errorMessage={meta.touched ? meta.error : undefined}
-                        disabled={isPending}
-                      />
-                    )}
-                  </Field>
+                  <TextInput
```
offerijns, line 361: I think we should just remove it, not really relevant information if it cannot be changed anyways.

offerijns (approved):

PR #1298, opened 2023-04-06
GHA Trigger for catalyst

gpmayorga:
Fixing tagging trigger for onboarding API in GHA and logic on how to detect the deployment should be catalyst


gpmayorga:
@sophialittlejohn  Make sure these values are the ones we want for Catalyst:
https://github.com/centrifuge/apps/blob/main/centrifuge-app/.env-config/.env.catalyst

sophialittlejohn (approved):
💯 🙏

PR #1314, opened 2023-04-17
fetch for generic subdoc link

jpangelle:
### Description

This pull request temporarily configures cent app to fetch for a generic subscription agreement from the Google bucket. Right now, all of the pool/tranche specific documents are identical anyway. Until we are ready to release the app or until we automate the subscription agreement upload step, then I don't think we are ready to have pool/tranche specific subscription agreements. It is just making it unnecessarily painful for testing.

### Approvals

- [ ] Dev

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -25,7 +25,10 @@ It's also recommended to run Prettier automatically in your editor, e.g. using [
 
 ### Onboarding API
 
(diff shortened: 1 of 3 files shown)


sophialittlejohn (approved):
🙏

PR #1327, opened 2023-04-28
EVM Wallets on Substrate

onnovisser:
### Description

Adds basic EVM support, behind a debug flag. Left todo is the option to have separate signing and sending of transactions, which is needed for creating loans. 
Also upgrades to WalletConnect V2

### Approvals

- [ ] Dev


centrifuge-js/src/utils/index.ts
```diff
@@ -141,3 +141,8 @@ export function computeMultisig(multisig: Multisig): ComputedMultisig {
     threshold: multisig.threshold,
   }
 }
+
+export function evmToSubstrateAddress(address: string) {
```
branan, line 145: We should probably make this an RPC/runtime API so that we can update the algorithm more easily

branan:
I'm really impressed how easily this all came together. `wrapSubstrateEvmSignAndSend` looks pretty much how I would expect it to. The rest of the frontend bits (especially the wallet component) are a bit out of my area of expertise to comment on

centrifuge-react/src/components/WalletProvider/WalletProvider.tsx
```diff
@@ -100,16 +135,44 @@ export function WalletProvider({
...
+  const consts = useCentrifugeConsts()
+  const centEvmChainId = useCentEvmChainId()
+
+  const evmChains = React.useMemo(() => {
+    const chains = {
+      ...evmChainsProp,
+    }
+    if (centEvmChainId) {
+      chains[centEvmChainId] = {
+        urls: ['https://fullnode.development.cntrfg.com/'],
```
sophialittlejohn, line 152: this should be passed down as an env variable

sophialittlejohn:
nice job @onnovisser!

I managed to get some DEVEL and aUSD into my evm wallet. From the screenshot it appears I have enough of both DEVEL and aUSD to invest.

Screenshot 2023-06-29 at 16 36 33

The tx fails though stating that my balance is too low 

Screenshot 2023-06-29 at 16 51 57

sophialittlejohn (approved):
🥳

PR #1328, opened 2023-05-01
Store encrypted files on IPFS and link to POD attributes

### LINKED ISSUE #1438: Store encrypted files on IPFS and link to POD attributes
## Background
When creating an asset the user wants to upload a file on top of entering all the asset details that are needed to tokenize the asset. 

## Requirements
* Add the option to upload a pdf file when creating an asset
* Currently it’s pinning the file as a JSON file that’s basically { contents: base64encodedcontents }. 
* In the future store it as a normal file which contents just are base64encodedcontents (Jeroen ran into issues with that)
> Jeroen: I think the decoding was failing because it couldn’t infer the content type, or something like that. The advantage would be that it would be more generic: external files that don’t follow this specific JSON format could more easily be used.
* When properly reviewed (SR Labs?) move the encryption methods into centrifuge.js.

hieronx:
### Description

This pull request adds an `encrypted-file` attribute type to loans. When a file is added, this is encrypted using a randomly generated key, pinned on IPFS, and both the IFPS hash and the key are stored in the POD document. This way anyone with access to the POD document can download the file and decrypt it.

Closes #1438 

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product


centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
```diff
@@ -338,7 +356,7 @@ function IssuerCreateLoan() {
           />
           {isLoggedIn ? (
             <>
-              <PageSection titleAddition="Select a template to enter the asset details.">
```
offerijns, line 341: This is unrelated to this PR but I noticed this line should have been removed already

centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
```diff
@@ -137,6 +138,23 @@ function TemplateField({ label, name, input }: TemplateFieldProps) {
           max={input.max}
         />
       )
+    case 'encrypted-file':
```
offerijns, line 141: We can consider also adding a `file` type, which is not encrypted and used as a public field. Or even just having only a `file` type and encrypting or not depending on whether the section is public.

centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
```diff
@@ -421,58 +439,149 @@ function IssuerCreateLoan() {
...
-              {
-                type: 'string',
-                value: String(value),
-              },
-            ]
-        }
-      })
-    )
+    (
+      await Promise.all(
```
offerijns, line 449: This had to be async to allow the encrypting and pinning within this method.

centrifuge-app/src/pages/IssuerPool/Assets/CreateLoan.tsx
```diff
@@ -421,58 +439,149 @@ function IssuerCreateLoan() {
...
+              })
+            )
+        )
+      )
+    ).flat()
   )
 }
+
+// Source: https://gist.github.com/chrisveness/43bcda93af9f646d083fad678071b90a
+async function aesGcmEncrypt(plaintext: string, password: string) {
```
offerijns, line 542: This of course should be moved somewhere else. And maybe replaced by a higher level library that does this.

centrifuge-app/src/types/index.ts
```diff
@@ -6,8 +6,8 @@ export type LoanTemplateAttribute = {
   label: string
   type: {
     primitive: 'string' | 'number'
-    statistics: 'categorical' | 'continuous' | 'ordinal' | 'descrete'
-    constructor: 'String' | 'Date' | 'Number'
+    statistics?: 'categorical' | 'continuous' | 'ordinal' | 'discrete'
```
offerijns, line 9: 1) Small typo here
2) I think this should be optional. It doesn't make sense to specify statistics for file uploads, and there will likely be other input types in the future for which there's no possible statistics.

centrifuge-js/src/modules/metadata.ts
```diff
@@ -2,12 +2,20 @@ import { first, from, map, Observable } from 'rxjs'
...
+    return inst.getFileObservable<T>(url)
+  }
+
   function getMetadata<T = any>(uri: string): Observable<T | T[] | null> {
     const url = parseMetadataUrl(uri)
     if (!url) {
       return from([])
     }
-    return inst.getMetadataObservable<T>(url)
+    return inst.getJsonObservable<T>(url)
```
offerijns, line 18: I think this all can be use a little refactoring. `getMetadata` doesn't seem specific to metadata, works for any file. `getMetadataObservable` really is a getter specifically for JSON files.

PR #1330, opened 2023-05-03
Refactor/liquidity charts data grouped by day

Hornebom:
### Description
Refactors Liquidity Charts to use data group by day instead of grouped by epoch

### Approvals

- [ ] Dev


annamehr:
Looks good!

PR #1341, opened 2023-05-08
check for matching email address

jpangelle:
### Description

This pull request enforces that when a user is verifying their email address, the email address in the database should match the email address in the jwt.

### Approvals

- [x] Dev


sophialittlejohn (approved):
💯

Issue #1343, opened 2023-05-09
Bug fixes for Tinlake in Centrifuge App


annamehr:
- [x] The total value locked and value locked and capacity of Tinlake pools seems to be off, see https://app-catalyst.k-f.dev/pools
- [x] It looks like the token price charts are saying “Data unavailable”. Can we just hide this?
![Screenshot 2023-05-09 at 10 31 49](https://user-images.githubusercontent.com/31280003/237040509-7860619c-c123-4528-a812-d354ab767400.png)

- [x] When I cancelled a tx manually in Metamask, the toast got stuck (I can’t close it). After reloading my account got disconnected and had to clear my local storage and then I was able to connect my wallet again. 
![Screenshot 2023-05-09 at 10 36 13](https://user-images.githubusercontent.com/31280003/237041681-d82e8ba4-9d07-40ad-b1df-a156f8efb18e.png)

- [x] When trying to redeem, the button just says “Approve BT1TIN” and then the input -> Better UX experience is to show the input first, and then chain Approve + Transfer afterwards 

- [x] ETH logos in wallet component are broken in production environments <img width="1512" alt="Screenshot 2023-05-11 at 16 40 29" src="https://github.com/centrifuge/apps/assets/24418964/2d18d019-05dc-4e64-99d4-355f9de82725">

Hornebom:
Hide PriceYieldChart fallback 'No data available' solved within https://github.com/centrifuge/apps/pull/1356

PR #1356, opened 2023-05-10
Hide PriceYieldChart fallback if no data available

Hornebom:
### Description
Subtask from https://github.com/centrifuge/apps/issues/1343

- hide the fallback message 'No data available'
- extracts the token card into dedicated component to handle data loaded state
- render 'default' variant of InteractiveCard as expand functionality is not needed

### Approvals

- [ ] Dev
- [ ] Product


annamehr:
Looks good!

annamehr (approved):

fabric/src/components/InteractiveCard/index.tsx
```diff
@@ -32,10 +32,11 @@ export const InteractiveCard: React.FC<InteractiveCardProps> = ({
   subtitle,
   children,
   onClick,
+  isOpen,
   ...rest
 }) => {
   const [hovered, setHovered] = React.useState(false)
-  const [open, setOpen] = React.useState(rest.isOpen ?? false)
+  const [open, setOpen] = React.useState(isOpen ?? false)
```
onnovisser, line 39: there's a `useControlledState` hook for this purpose
```js
const {isOpen: externalOpen, setOpen: setExternalOpen} = props

const [open, setOpen] = useControlledState(false, externalOpen, setExternalOpen)
```
it will need the external setOpen so the component controlling the state can set it's own state. Otherwise the internal setOpen has no effect, when isOpen is passed

fabric/src/components/InteractiveCard/index.tsx
```diff
@@ -19,28 +20,33 @@ type OwnProps = {
...
+  const {
+    variant = 'default',
+    icon,
+    title,
+    titleAddition,
+    secondaryHeader,
+    subtitle,
+    children,
+    onClick,
+    isOpen = false,
```
onnovisser, line 38: these shouldn't have defaults

onnovisser (approved):

PR #1372, opened 2023-05-12
Refactor/Review calculation of value locked and capacity of pool

Hornebom:
### Description

Subtask from: https://github.com/centrifuge/apps/issues/1343

### Approvals
- [ ] Dev
- [ ] Product


centrifuge-app/src/components/PoolCard.tsx
```diff
@@ -113,16 +111,22 @@ export function PoolCard({ pool }: PoolCardProps) {
         />
         <LabelValueStack
           label="Capacity"
-          value={
```
onnovisser, line 116: this should be correct. I just realized we only need to re-enable the `currentNAV` call in `useTinlakePools`, which we disabled when the NAV was bugged.

Hornebom: Updated 👍

onnovisser (approved):

PR #1374, opened 2023-05-12
fix: kyc decline

jpangelle:
### Description

This pull request fixes the issue where a user still proceeds in the onboarding flow even when their KYC submission was denied.

### Approvals

- [ ] Dev

### Screenshots

The user will see this screen and will no longer be advanced to the next step:
image


sophialittlejohn (approved):
💯

PR #1375, opened 2023-05-15
fix: autoconnect wallet when visiting through gnosis safe app

jpangelle:
### Description

This pull request autoconnects the wallet when visiting through the Gnosis Safe app

### Approvals

- [x] Dev


onnovisser (approved):
🎉

Issue #1199, opened 2023-02-16
Reporting tab v2


annamehr:
[Figma file](https://www.figma.com/file/9iwvzHuvbjlHXRWlPWRxxc/Centrifuge-app---Version-2-(Branch)?node-id=7240%3A54828&t=wXEmwuLh6REu7qVM-0)

## Background
As an investor I want to see how the pool performs and be able to download reports as a csv file. 
As the first step Ben implemented `investor transactions`, `pool balance` and `asset list`. 

## Next

- [ ] Code refactoring
- [ ] Design implementation
- [ ] Add a `Borrower transactions report` and option to download the report as a csv file 

The schema that the subquery exposes for BorrowerTransaction is here: https://github.com/centrifuge/pools-subql/blob/main/schema.graphql#L233

Columns would be:
Loan ID
Epoch #
Timestamp
Type (created, priced, financed, repaid, closed)
Amount

Pdf download can be dismissed for now.

PR #1384, opened 2023-05-16
Feat/implement design reporting tab

Hornebom:
### Description
**Note**
Working pull request with verified commits: https://github.com/centrifuge/apps/pull/1520
This pr is only for reference

Issue: https://github.com/centrifuge/apps/issues/1199

### Approvals
- [ ] Dev
- [ ] Product


annamehr:
Some feedback: 

Pool balance:
* should we not display group by month if date range is last week and instead only display day in the dropdown?
* should we display the values on the left side instead of on the right side and have this big white spacing in between?

Screenshot 2023-05-17 at 12 53 00

offerijns:
A few testing notes:
- I wonder if it would make sense to fix the row names (the first column of the data table) and just scroll the remaining columns. If you look on https://app-pr1384.k-f.dev/pools/3317493590/reporting and scroll right, the row names (Pool value, Asset value, ...) quickly disappear.
- The data table width seems to be limited to the initial screen width and not grow with the full scrollable width.
Screenshot 2023-05-19 at 12 06 17

- I think we should parse/re-format the borrower and investor tx types a bit.

For the borrower tx type, I would suggest to just capitalize the first letter. E.g. `CREATED` => `Created`, `REPAID` => `Repaid`. And change `BORROWED` to `Financed` (as we consistently use this naming in the UI, while borrowing is just a term used in the code).

For the investor tx type, I would suggest:
`INVEST_ORDER_UPDATE` => `Investment order updated`
`REDEEM_ORDER_UPDATE` => `Redemption order updated`
`INVEST_ORDER_UPDATE` if currency amount is 0 => `Investment order cancelled`
`REDEEM_ORDER_UPDATE` if currency amount is 0 => `Redemption order cancelled`
`INVEST_EXECUTION` => `Investment executed`
`REDEEM_EXECUTION` => `Redemption executed`
`INVEST_COLLECT` => `[TRANCHE_TOKEN_SYMBOL] received in wallet` (e.g. `NS2DRP received in wallet`)
`REDEEM_COLLECT` => `[POOL_CURRENCY_SYMBOL] received in wallet` (e.g. `USDT received in wallet`)
`TRANSFER_IN` => `Deposited [TRANCHE_TOKEN_SYMBOL]`
`TRANSFER_OUT` => `Withdrawn [TRANCHE_TOKEN_SYMBOL]`
     
- The `Loan` column of the `Borrower transactions` view should be renamed to `Asset ID`, and should just display the number, not the pool ID prefix and the `-`. Similarly, the epoch column data should just be the epoch # (1, 2, 3, ...), not include the pool id prefix.
- The `Token amount` column of the Borrower transactions view should be `[POOL_CURRENCY_SYMBOL] amount` (e.g. USDT amount)

gpmayorga:
For some reason this PR does not trigger Github Actions correctly 🤷‍♂️

annamehr:
My testing feedback.. it already looks a lot better! @Hornebom 

On the first screenshot:
* It looks like the reporting data for the US treasuries pool is not synced and therefore not displayed. The reserve of BT test pool is 70k USDT but displayed is 0.0 USDT and Value locked is 1,006,888 USDT and displayed is only 0.0 USDT. 
Screenshot 2023-07-10 at 12 15 30

Second screenshot: 
* I want to see the pool balance of every day last week but I can only see July 6th, looks like the group by filter isn't working yet.  
Screenshot 2023-07-10 at 12 17 39

fabric/src/components/DateRange/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const date = new Date($date)
+    return new Date(date.setFullYear(date.getFullYear() - 1))
+  },
+} as const
+
+export function DateRange({ onSelection, defaultOption = rangeOptions[1], start, end }: DateRangeProps) {
+  const [value, setValue] = React.useState(defaultOption.value)
+  const [startDate, setStartDate] = React.useState(start || new Date())
+
+  React.useEffect(() => {
```
onnovisser, line 52: this can just be in the state initializer
```
const [startDate, setStartDate] = React.useState(getDate[defaultOption.value](end))
```

centrifuge-app/src/components/DataTable.tsx
```diff
@@ -68,6 +69,7 @@ export const DataTable = <T extends Record<string, any>>({
   )
 
   const [currentSortKey, setCurrentSortKey] = React.useState(defaultSortKey || '')
+  const [setNode, { scrollWidth }] = useElementScrollSize()
```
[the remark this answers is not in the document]
Hornebom: refactored

fabric/src/components/DateRange/index.tsx
Hornebom: refactored

onnovisser (approved):

Issue #1386, opened 2023-05-16
Testing feedback 


annamehr:
- [ ] Change the TIN APY and remove 0.00% and put `Variable` as the rate with a tooltip `Junior tranche accrues variable returns, which are based on the spread of average financing fee of the assets in the loans and the fixed interest rate that the senior tranche is accruing.`

<img width="701" alt="Screenshot 2023-05-16 at 18 57 20" src="https://github.com/centrifuge/apps/assets/31280003/879a8638-8b45-4f76-bb3d-21c257b6eb66">

- [ ] Remove the pool account button for Ethereum pools

- [ ] Align the external link icons in the Governance menu item with the chevron above as shown in the [designs](https://www.figma.com/file/Yl5BMnm3AOlxIt0g41PbcT/Centrifuge-Nav?type=design&node-id=111-10648&t=qgUGS5VXVlYidIhB-0)  

<img width="187" alt="Screenshot 2023-05-16 at 19 10 08" src="https://github.com/centrifuge/apps/assets/31280003/bea8be6c-96b4-4589-b5ac-353746ca0dca">

onnovisser:
I can look at nr. 4 and 6

Hornebom:
5: https://github.com/centrifuge/apps/pull/1389
7: https://github.com/centrifuge/apps/pull/1390

Will pick up 1,2 and 3

annamehr:
🔥🔥🔥🔥🔥

PR #1395, opened 2023-05-18
Centrifuge App: Show right token in invest form

onnovisser:
### Description

- Show right token in invest form
- Round low balances to 2 decimals in wallet button

Bug 4 and 6 of #1386

### Approvals

- [ ] Dev


annamehr:
Looks good!

PR #1433, opened 2023-06-02
Fix Asset Lists

AStox:
### Description

This pull request fixes the broken asset lists by switching the subgraph query to get loans via pools instead of querying all loans and filtering by pools. For some reason the latter is incredibly slow. Affected pools are NS2 and Branch 3, but this should speed up asset list querying across the board.

### DIFF
--- a/centrifuge-app/.env-config/.env.development
+++ b/centrifuge-app/.env-config/.env.development
@@ -11,6 +11,6 @@ REACT_APP_POOL_CREATION_TYPE=immediate
 REACT_APP_RELAY_WSS_URL=wss://fullnode-relay.development.cntrfg.com
 REACT_APP_SUBQUERY_URL=https://api.subquery.network/sq/centrifuge/pools-development
 REACT_APP_SUBSCAN_URL=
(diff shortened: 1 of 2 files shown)


centrifuge-app/.env-config/.env.development
```diff
@@ -11,6 +11,6 @@ REACT_APP_POOL_CREATION_TYPE=immediate
 REACT_APP_RELAY_WSS_URL=wss://fullnode-relay.development.cntrfg.com
 REACT_APP_SUBQUERY_URL=https://api.subquery.network/sq/centrifuge/pools-development
 REACT_APP_SUBSCAN_URL=
-REACT_APP_TINLAKE_NETWORK=goerli
+REACT_APP_TINLAKE_NETWORK=mainnet
```
jpangelle, line 14: Can you revert this change?

Instead you would copy the contents of `.env.development` to a new file called `.env.development.local` that would be `gitignored`.

jpangelle: Ah ok, cool!

Issue #1443, opened 2023-06-12
Display asset value for each pool


annamehr:
As a user I want to see the total value of all assets in the pool. 

* Display Asset value
* Asset value is calculated by summarising the value of all ongoing assets in the pool. Similar to how it was implemented on Tinlake (https://legacy.tinlake.centrifuge.io/pool/0x53b2d22d07E069a3b132BfeaaD275b10273d381E/new-silver-2/assets) 
* Tooltip: Asset value is the net value of all assets being financed in the pool. 
<img width="500" alt="Screenshot 2023-06-12 at 11 56 16" src="https://github.com/centrifuge/apps/assets/31280003/99ab3990-2fc4-44e7-9f80-d31d946ce30b">
* Remove average maturity from the asset list of all BT pools (B1-B4)

PR #1444, opened 2023-06-12
Pool asset value

Hornebom:
### Description
- Asset value is calculated by summarising the value of all ongoing assets in the pool
- removes Average maturity

Related issue: https://github.com/centrifuge/apps/issues/1443

### Approvals

- [ ] Dev
- [ ] Product

### Screenshots
Screenshot 2023-06-12 at 13 56 30


onnovisser (approved):

offerijns:
It's now displaying the sum of all loan debts, which is different from the asset value. That should be `ASSESSOR.currentNAV` or something like that for Tinlake pools (you can look up how this value was retrieved in Tinlake), and the `portfolioValuation` for Centrifuge pools.

Hornebom:
@offerijns in tinlake it's `poolData.netAssetValue` what have become `pool.nav` for both tinlake and centrifuge pools. Thanks for checking in

PR #1454, opened 2023-06-13
CentrifugeApp: Fix aUSD display

sophialittlejohn:
### Description

The invest/redeem component was unable to display aUSD. Seems like the currency key may have changed.

### Approvals

- [ ] Dev


centrifuge-js/src/modules/pools.ts
```diff
@@ -2283,11 +2283,17 @@ function getEpochStatus(epochExecution: Pick<EpochExecutionData, 'challengePerio
...
 }
 
-export function findCurrency<T extends Pick<CurrencyMetadata, 'key'>>(
+export function findCurrency<T extends Pick<CurrencyMetadata, 'key' | 'symbol'>>(
   currencies: T[],
   key: CurrencyKey
 ): T | undefined {
-  return currencies.find((currency) => looksLike(currency.key, key))
+  const curr = currencies.find((currency) => {
+    if (typeof key === 'string') {
```
sophialittlejohn, line 2291: if an account had `aUSD` it was still displaying as zero. The key used to find aUSD before the change was `{ foreignAsset: 1 }` which caused the find to return undefined.

onnovisser (approved):

PR #1455, opened 2023-06-14
Centrifuge App: Fix tranche updates

onnovisser:
### Description

- Fixes "Epoch and Tranches" form submission
- Disables use of `useCentrifugeQueries`, which occasionally entered an infinite loop
- Fixes negative transfer when removing hot wallet from AO
- Fixes pending multisigs not refetching after approving

### Approvals

- [ ] Dev


sophialittlejohn (approved):
nice job fixing 🧹

PR #1461, opened 2023-06-15
CentrifugeJS: Rename pod.url to pod.node

onnovisser:
### Description

Rename `pod.url` to `pod.node` in pool metadata and add `pod.indexer`. Important to get this in first, in case a pool launches before the indexer work is done, so we don't have to migrate metadata later.

### Approvals

- [ ] Dev


offerijns (approved):
🙌

PR #1463, opened 2023-06-15
Faucet: Use foreignAsset key for faucet & Onboarding: minimum age 18

### LINKED ISSUE #1456: Disable entering same email address twice + check birthdate
Environment: Firefox + Chrome, Mac

#### Steps to reproduce
* Add new email address in step 3 
* In step 4 change new email address to already used email address

#### Expected Result
* Show an error that email address is already registered 
#### Actual Result
* User was able to use same email address

## Additional steps
* Check in step 3 if email address already exists and if not allow next proceeding to next step
* Disallow onboarding if user is younger than today - 18 years

Screenshot 2023-06-14 at 11 43 07

sophialittlejohn:
### Description

This pull request...
- fixes aUSD transfer from the faucet (it was using an outdated version of aUSD)
- makes the faucet agnostic to pool currencies
- moves the faucet to display only in pool detail pages and sends the native currency and the pool currency to the user

- enforces a minimum age of 18 to onboard (also for UBOs). Closes #1456 

### Approvals

- [ ] Dev


jpangelle (approved):
looks good!

PR #1483, opened 2023-07-01
Fix broken development environment

Hornebom:
### Description
- Unexpected token error thrown in graphql-request caused by vite

Context and fix:
[Vite Environment Variable Replacements #13594](https://github.com/vitejs/vite/discussions/13594)
https://github.com/vitejs/vite/discussions/13594#discussioncomment-6250699

### Approvals

- [ ] Dev

### Screenshots
Screenshot 2023-06-30 at 18 32 10

### DIFF
--- a/centrifuge-app/src/utils/tinlake/useTinlakePools.ts
+++ b/centrifuge-app/src/utils/tinlake/useTinlakePools.ts
@@ -11,7 +11,6 @@ import {
 import { useCentrifuge } from '@centrifuge/centrifuge-react'
 import { BigNumber } from '@ethersproject/bignumber'
 import BN from 'bn.js'
-import { request } from 'graphql-request'
 import * as React from 'react'
 import { useQuery } from 'react-query'
 import { lastValueFrom } from 'rxjs'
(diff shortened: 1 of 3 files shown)


onnovisser:
I think we can just do the one call it's used for manually, like the other calls, and ditch the graphql and graphql-request libs, and save 50kb

centrifuge-app/src/utils/tinlake/useTinlakePools.ts
```diff
@@ -237,34 +236,58 @@ function getTinlakeLoanStatus(loan: TinlakeLoanData) {
...
+        },
+      }),
+    })
 
-  const data = await request<{ data: any[] }>('https://graph.centrifuge.io/tinlake', query)
+    if (response?.ok) {
+      const { data } = await response.json()
+      pools = data.pools
+    } else {
+      console.warn(`Issue fetching loans for Tinlake pool ${poolId}. Status:`, response?.status)
```
onnovisser, line 284: I would remove the try/catch and throw here instead. then react query will try again

onnovisser (approved):

PR #1486, opened 2023-07-04
Update .env.development with new POD

gpmayorga:
New pod deployment for DEV
In order to close this PR we should test the endpoint works out (there has been some networking changes) I'll then open a PR with the endpoints for the rest of the environments


offerijns (approved):
Works for me 👍

Issue #1497, opened 2023-07-10
Wallet selector (EVM compatible accounts)


annamehr:
🌀 [Figma Design](https://www.figma.com/file/9iwvzHuvbjlHXRWlPWRxxc/Centrifuge-app---Version-2-(Branch)?type=design&node-id=7888-50405&mode=design&t=RCWdl8FLyzY8AXI4-0)

## Background 
For Centrifuge Chain being EVM compatible, there are some changes in the wallet selector needed for the user to send CFG tokens to the correct Substrate address and to choose between substrate and evm address. 

## Flow 
<img width="654" alt="Screenshot 2023-07-11 at 09 45 44" src="https://github.com/centrifuge/apps/assets/31280003/d9b9c8b4-2b17-4860-9a42-e2a8f40ebab1">

## Requirements
* Remove Network/Wallet illustration at the top 
<img width="383" alt="Screenshot 2023-07-10 at 18 28 47" src="https://github.com/centrifuge/apps/assets/31280003/f5fcc472-ba86-4161-8c27-6aab8e558322">

* When Centrifuge Chain is chosen, show Substrate account and the option to connect to the EVM account 
* When connected clicking on `Switch account ` leads to account page that includes network and wallet 
<img width="360" alt="Screenshot 2023-07-11 at 09 46 36" src="https://github.com/centrifuge/apps/assets/31280003/35a3dd38-a738-48de-be6d-f56590e19afc">

* Show substrate address in wallet
<img width="214" alt="Screenshot 2023-07-10 at 18 35 25" src="https://github.com/centrifuge/apps/assets/31280003/e83193e7-b1c4-429d-8e98-65e5905beffc">

PR #1498, opened 2023-07-13
Centrifuge App: Wallet dialog and wallet menu changes

onnovisser:
### Description

#1497

### Approvals

- [ ] Dev

### Screenshots

image

image


annamehr:
@onnovisser looks really good. One thing I'm missing is the Use EVM account button. I enabled ShowEVMonSubstrate and connected with Talisman but I couldn't switch to the EVM address. Is there a better way how I can test this feature? 
Screenshot 2023-07-14 at 17 15 15

centrifuge-react/src/components/WalletProvider/utils.ts
```diff
@@ -30,6 +31,19 @@ export function useNetworkName(network: Network) {
...
+    evm,
+    substrate: { evmChainId },
+  } = useWallet()
+  return (network: Network) =>
+    network === 'centrifuge' || network === evmChainId ? centrifugeLogo : evm.chains[network]?.iconUrl ?? ''
+}
+
+export function useNetworkIcon(network: Network) {
+  return useGetNetworkIcon()(network)
+}
```
sophialittlejohn, line 45: wondering if these two should be combined for simplicity?

sophialittlejohn (approved):

centrifuge-react/src/components/WalletProvider/utils.ts
onnovisser: we could get rid of the second one, but it's also kinda convenient

PR #1501, opened 2023-07-20
feat: Navigation component

Hornebom:
### Description
- add navigation component and standalone sub components (mainly to be able to extend the NavigationItem as Link in centrifuge-app)

### Approvals

- [ ] Dev


fabric/src/components/Navigation/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export type NavigationProps = {
+  items: {
+    label: string
+    href: string
+    isActive: boolean
+  }[]
+}
+
+export function Navigation({ items }: NavigationProps) {
```
onnovisser, line 13: I feel the naming is quite generic for such a specific kind of navigation. maybe something like SegmentedControls, SelectionControls, SelectionNavigation, or something

PR #1513, opened 2023-07-28
Pool overview filter

Hornebom:
### Description

As subtask of

- functional implementation of pool filter
- filter based on query params to enable shareable urls and a light weight option to store user selection (navigate from overview to pool detail and use browser history back will have filter settings preserved)

### Approvals
- [ ] Dev
- [ ] Designer


centrifuge-app/src/components/PoolFilter/FilterMenu.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { SearchKeys } from './types'
+import { toKebabCase } from './utils'
+
+export type FilterMenuProps = {
+  label: string
+  options: string[]
+  searchKey: SearchKeys['ASSET_CLASS' | 'POOL_STATUS']
+}
+
+export function FilterMenu({ label, options, searchKey }: FilterMenuProps) {
```
onnovisser, line 14: Nice!

centrifuge-app/src/pages/Pools.tsx
```diff
@@ -16,19 +27,73 @@ export function PoolsPage() {
...
+      )}
+    </Stack>
   )
 }
+
+async function formatPoolsData(pools: (Pool | TinlakePool)[], cent: Centrifuge): Promise<PoolCardProps[]> {
+  const promises = pools.map(async (pool) => {
+    const tinlakePool = pool.id?.startsWith('0x') && (pool as TinlakePool)
+    const mostSeniorTranche = pool?.tranches?.slice(1).at(-1)
+    const metaData = typeof pool.metadata === 'string' ? await metadataQueryFn(pool.metadata, cent) : pool.metadata
```
onnovisser, line 78: this shouldn't fetch metadata on its own. metadata is already being fetched via useQuery and this ignores all caching and will fetch it again. Instead of the asyncMemo, you can have something like
```
const centPools = usePools()
const poolMetas = useMetadataMulti<PoolMetadata>(centPools?.map((p) => p.metadata) ?? [])
const poolMetasById = ...

const pools = formatPoolsData(...)
```

annamehr:
@Hornebom When clicking on tokens and then back to pools I receive this error message `No results found with these filters. Try different filters.` Same error appears when clicking on a pool and then returning to the pools overview.

annamehr:
@Hornebom 
- Connect button is larger than in the designs and could be smaller
- There is quite a lot of spacing between Pools content and the left column --> possible to reduce the spacing here?
- quite a big distance between asset class and value locked in the table compared to the designs
- What about the call to action which is in line with the tvl graph? It could lead to the global onboarding flow.
Pools on Centrifuge let investors earn yield from real-world assets. `Get started button` leads to global onboarding flow
- Tooltips are missing for asset type, value locked and pool status

centrifuge-app/src/pages/Pools.tsx
```diff
@@ -31,9 +33,15 @@ function Pools() {
...
-  const pools = useAsyncMemo(async () => {
-    return !!listedPools?.length ? await formatPoolsData(listedPools, cent) : []
-  }, [listedPools.length])
+  const centPools = listedPools.filter(({ id }) => !id.startsWith('0x')) as Pool[]
+  const centPoolsMetaData: PoolMetaDataPartial[] = useMetadataMulti<PoolMetadata>(
+    centPools?.map((p) => p.metadata) ?? []
+  ).map((q) => q.data)
+  const centPoolsMetaDataById = getMetasById(centPools, centPoolsMetaData)
+
+  const pools = React.useMemo(() => {
```
onnovisser, line 42: useMemo will invalidate every render because `centPoolsMetaDataById` is a new object every render. The useMemo below as well. I don't think the memos are needed. the stuff inside isn't too heavy

onnovisser (approved):

PR #1539, opened 2023-08-18
OnboardingAPI: liquidity pool authentication and remark signing

sophialittlejohn:
### Description

This pull request...
- implements auth for base (adds chain id to auth token)
- implements remark signing for base (checks which network the remark was signed on and validates there)
- adds base network behind debug flag
- adds buttons in UI to onboard to base and avalanche - doesn't include proper integration of base and avalanche in UI since it's out of scope of this task

You can use this faucet to get base goerli eth: https://faucet.quicknode.com/drip

Closes #<issue-number>

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product


centrifuge-app/src/components/InvestRedeem/InvestRedeem.tsx
```diff
@@ -64,7 +64,7 @@ type Props = {
 // @ts-ignore
 const listFormatter = new Intl.ListFormat('en')
 
-export function InvestRedeem({ networks = ['centrifuge'], ...rest }: Props) {
+export function InvestRedeem({ networks = ['centrifuge', 43114, 43113, 8453, 84531], ...rest }: Props) {
```
jpangelle, line 67: Are these the chain id's for Base?

centrifuge-app/src/components/Root.tsx
```diff
@@ -97,6 +101,14 @@ const evmChains: EvmChains =
           urls: [`https://goerli.infura.io/v3/${infuraKey}`],
           iconUrl: goerliLogo,
         },
+        8453: {
+          urls: ['https://mainnet.base.org'],
+          iconUrl: 'https://docs.base.org/img/logo.svg',
+        },
```
jpangelle, line 107: Why is the `mainnet` base id in the `else` of the ternary? It looks like the ternary is checking if the network is `mainnet`.

centrifuge-app/src/components/InvestRedeem/InvestRedeem.tsx
sophialittlejohn: they're the chain id's for avalanche and base. Let me try and clean it up a bit so it's more clear

centrifuge-app/src/components/Root.tsx
sophialittlejohn: it's just so that we can use mainnets on dev as well

centrifuge-react/src/components/WalletProvider/evm/chains.ts
```diff
@@ -53,6 +53,31 @@ const chainExtendedInfo = {
...
+  36: {
+    name: 'Centrifuge',
+    nativeCurrency: { name: 'Altair', symbol: 'AIR', decimals: 18 },
+    blockExplorerUrl: 'https://www.subscan.io/',
+  },
+  2000: {
+    name: 'Centrifuge',
+    nativeCurrency: { name: 'Centrifuge', symbol: 'CFG', decimals: 18 },
+    blockExplorerUrl: 'https://www.subscan.io/',
+  },
```
sophialittlejohn, line 70: Added these for the network switch in the invest redeem component since it needs a name for centrifuge networks

centrifuge-app/src/components/OnboardingAuthProvider.tsx
```diff
@@ -232,7 +233,8 @@ Issued At: ${new Date().toISOString()}`
       signature: signedMessage,
       address,
       nonce,
-      network: evmChainId ? 'evmOnSubstrate' : 'evm',
+      network: evmChainId === 2000 ? 'evmOnSubstrate' : 'evm',
```
onnovisser, line 236: `useWallet()` also returns `isEvmOnSubstrate` and `substrate.evmChainId`. you can check against one of those instead, because the chainId isn't always 2000

centrifuge-app/src/components/Root.tsx
```diff
@@ -146,6 +160,7 @@ export function Root() {
                       setIsThemeToggled(!!state.alternativeTheme)
                       setShowAdvancedAccounts(!!state.showAdvancedAccounts)
                       setShowAvalanche(!!state.showAvalanche)
+                      setShowBase(!!state.showBase)
```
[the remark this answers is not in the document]
onnovisser: it's a typing issue with the debug flags. it thinks all flags return `boolean | string`

centrifuge-react/src/components/WalletProvider/evm/chains.ts
onnovisser: cent evm chain already gets defined in the WalletProvider, here: https://github.com/centrifuge/apps/blob/main/centrifuge-react/src/components/WalletProvider/WalletProvider.tsx#L156

onnovisser: the other 2 don't need to be there either, as the network switch checks if the network is `"centrifuge"` to show the centrifuge/altair network name

onboarding-api/src/utils/networks/evm.ts
```diff
@@ -0,0 +1,66 @@
...
+import { InferType } from 'yup'
+import { signAndSendDocumentsInput } from '../../controllers/emails/signAndSendDocuments'
+import { HttpError } from '../httpError'
+import RemarkerAbi from './abi/Remarker.abi.json'
+
+const getEvmProvider = (chainId: number | string): Provider => {
+  switch (chainId.toString()) {
+    case '1': // eth mainnet
+    case '5': // goerli
+    case '2000': // cent on evm
```
onnovisser, line 15: evm chain id will need to fetched from chain somewhere with `api.query.evmChainId.chainId()`

centrifuge-react/src/components/WalletProvider/evm/chains.ts
sophialittlejohn: good point, will remove the one with chain id 2000. For the others though, I was thinking it would be nicer to be able to fetch all network information the sam. if we don't include the cent chains in the chainExtendedInfo then we always have to check first if the network is centrifuge and handle it differently, even just to display the name

sophialittlejohn: hmm but now I see that this may be the wrong place to add the cent chain ids... will revert and suggest revisiting when your part is done as well!

onnovisser (approved):

PR #1563, opened 2023-09-06
Centrifuge App: Add debug flag for pool create type

onnovisser:
### Description

This pull request...

Closes #<issue-number>

### Approvals

- [ ] Dev


sophialittlejohn:
If the the createType is immediate I think we should hide the required deposit next to the cancel button since there is no deposit required as far as I know

onnovisser:
There's no proposal deposit, but there's still a deposit required for the pool itself

sophialittlejohn (approved):
🙌

PR #1580, opened 2023-09-14
ff-prod deploys (and less code)

gpmayorga:
### Description

To deploy a "preview" of the main branch code pointing to the prod endpoints.

Screenshot 2023-09-14 at 00 26 56
https://kflabs.slack.com/archives/C04GNE3JP6Z/p1683906262656619

Given the increasing number of "dual-deploys" I've rearranged things a little to allow to deploy to N environments on a given GH trigger.

Summary:
\+ group common workflow code
\+ universal "deploy all things" file
\+ Less lines of code at the end 😎 
\+ Dual prod (from main branch) without additional config files
\- main branch will now build all artifacts instead of selectively building depending on what folder changes

Check out the new main branch jobs deploying to 3 environments:
https://github.com/centrifuge/apps/actions/workflows/main-branch-deploys.yml
https://github.com/centrifuge/apps/actions/runs/6181041679

This pull request...
Saves some workflow code by grouping it and triggers deployments from main with prod config (app only)

Ah! And here's the preview:
https://app-ff-production.k-f.dev/pools


sophialittlejohn:
It looks like some of the inputs don't resolved properly both in the PR checks and in the [PR job](https://github.com/centrifuge/apps/actions/runs/6190415742/job/16806650017?pr=1580)

Screenshot 2023-09-14 at 17 53 08

gpmayorga:
A blip in the matrix, I think we're good

sophialittlejohn:
Will they fix themselves when this PR is merged?

gpmayorga:
Oh wait, I totally missed your point, you mean the ${{ inputs.environment }}
Since I'm setting up the name of the job dynamically, skipped jobs do not get to replace the variable for the actual runtime value.
In cases like `ff-prod` onboarding and pinning will never be deployed so in every build we will see those ${{

Pushing a fix to try and setup default suffixes when skipped.

gpmayorga:
> Will they fix themselves when this PR is merged?

Ok I tried a few options but it seems that since deploy_all.yml only trigger is being called by another workflow, the job names inside it will always be preceded by the parent workflow calling, so I don't think we really need unique names there.

I've disabled the naming and replaced for much simpler indicative names

gpmayorga:
Closed in favor of #1598

PR #1641, opened 2023-10-17
Centrifuge App: Portfolio asset allocation

onnovisser:
### Description

image

### Approvals

- [ ] Dev


sophialittlejohn:
nice job! Could you add tooltips when you hover on the chart?

sophialittlejohn:
nice! one more tiny request for the tooltips, could you add the color of the hovered item in the tooltip? Would be more consistent with the tooltips are already use for other charts
Screenshot 2023-10-19 at 14 44 10

onnovisser:
@sophialittlejohn 👍 added

sophialittlejohn (approved):
💯

PR #1659, opened 2023-10-24
Centrifuge App: Prime overview page

onnovisser:
### Description

Adds the Prime overview page. Also adds a FilterableTableHeader component for filtering table data. Will need an update to fetch the portfolio values via the new Runtime API when it's available.

### Approvals

- [ ] Dev

### Screenshots

image


centrifuge-app/src/components/Menu/index.tsx
```diff
@@ -53,6 +54,12 @@ export function Menu() {
           Portfolio
         </PageLink>
       )}
+      {showPrime && address && (
```
sophialittlejohn, line 57: I think we should be able to show prime whether an address is connected or not

sophialittlejohn (approved):
💯

centrifuge-app/src/components/Menu/index.tsx
onnovisser: yes, good point

PR #1660, opened 2023-10-24
Centrifuge App: Code splitting and cleaning up unused files

onnovisser:
### Description

Removing some old unused files, splitting out a few things that aren't needed for initial page load, like pages other than the pools page, charts and the pdf viewer. Overall cutting the initially loaded bundle by about 30%.

### Approvals

- [ ] Dev


sophialittlejohn (approved):
Love it 💯

PR #1663, opened 2023-10-25
Centrifuge App: Table redesign

onnovisser:
### Description

Implements new table design and updates tables in the app. Uses CSS Subgrid for better layout of columns. Chrome only recently added support for this, so users with Chrome <= 116, will get a fallback table, which will look a little off as the columns won't line up perfectly in some cases. Most Chrome users should be on version  >= 117 in a month or so, at which point we can remove the fallback.

### Approvals

- [x] Dev

### Screenshots

image

image

image

image

image

image

image


jpangelle (approved):
lgtm. nice work!

annamehr:
Found one styling issue otherwise looks good :) 

image

onnovisser:
fixed 🙂

onnovisser:
@annamehr ah I only fixed the color, I see the font-size was off too. will fix

Issue #1593, opened 2023-09-25
Portfolio page: Overview graph


sophialittlejohn:
[Specs](https://centrifuge.hackmd.io/hZcXZWLhQjqYHkwnUi6Svw)
[Design](https://www.figma.com/file/w5svs3Peu6obgsD79tMbQj/Portfolio-MVP---V1?type=design&node-id=1917-11453&mode=design&t=38r8EgzdxFF9bEdC-0)
![Screenshot 2023-10-30 at 15 42 34](https://github.com/centrifuge/apps/assets/31280003/5490a503-1db3-4f91-b1de-5a38a7751f78)

Current portfolio value (YTD, 30D, 90D, All).

**Calculations**
* Show the total amount of all investments (the portfolio) in USD
    * Show the total amount of all investments (the portfolio) in USD = 
`investment_portfolio(AccountId) -> [(PoolId, CurrencyId, TrancheId, Balance)]`
     * The front-end should call once this method, and for each `CurrencyId` found, ask `OrmlAssetRegistry` to get the correct decimals to make the correct sum. The expected amount of different `CurrencyId` is small: only the stable coins of pools.
    * By now, we can assume all currencies are stable coins for USD, and all stable coins are 1:1 with USD. In the future, the UI should display a list with different fiat currencies or even split stables. 
    * Default filter YTD

Parent ticket: https://github.com/centrifuge/apps/issues/1525

PR #1670, opened 2023-10-30
CentrifugeApp: Portfolio page Overview graph

sophialittlejohn:
### Description

This pull request...
- adds UI for pool overview graph with mocked data

Still missing:
- UI update to new designs
- read data from subql
- time frame tabs

#1593

Closes #<issue-number>

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product


TimHoub:
Some minor UI improvements that are a not in line with the design yet.
- The 'USD' on the left side looks a bit weird and is not in the designs.
- The active state on the filter is a little bit different. -> the non active items should have a font weight that is lighter than the active one. 
- The filter alignment should be on the same width as the right side of the graph, now it is bit too much to the right.

Apart from that, looks great!

TimHoub (approved):

jpangelle:
> Some minor UI improvements that are a not in line with the design yet.
> 
> - The 'USD' on the left side looks a bit weird and is not in the designs.
> 
> - The active state on the filter is a little bit different. -> the non active items should have a font weight that is lighter than the active one. 
> 
> - The filter alignment should be on the same width as the right side of the graph, now it is bit too much to the right.
> 
> 
> 
> Apart from that, looks great!

Regarding the USD bit, this was a suggestion from Dennis in order to signify what the y-axis represents. Is there a position we can place it that would look cleaner?

jpangelle:
@TimHoub @annamehr Your comments have been addressed.

centrifuge-app/src/components/DataTable.tsx
```diff
@@ -55,11 +55,13 @@ const sorter = <T extends Record<string, any>>(data: Array<T>, order: OrderBy, s
   if (order === 'asc') {
     return data.sort((a, b) => {
       if (sortKey === 'nftIdSortKey') return new BN(a[sortKey]).gt(new BN(b[sortKey])) ? 1 : -1
+      if (sortKey === 'position') return a[sortKey].gt(b[sortKey]) ? 1 : -1
```
onnovisser, line 58: this feels kinda weird to have in here

sophialittlejohn: I think the table sorting requires the sortable value to be a number, maybe you could adapt this so that the sorter checks first if the value is a Dec/BN and sort accordingly? Then you would probs be able to remove the line above it as well

centrifuge-app/src/components/Portfolio/usePortfolio.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  return useMemo(() => {
+    if (
+      dailyTrancheStatesByTrancheId &&
+      transactionsByTrancheId &&
+      Object.keys(dailyTrancheStatesByTrancheId).length === dailyTrancheStates?.length
+    ) {
+      const today = new Date()
+
+      return Array(rangeValue + 1)
```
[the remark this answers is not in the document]
jpangelle: Hm I think so but if the user toggles the selection to another range, it require another call to the subquery to get the data rather than having all of the data already. Might feel sluggish.

centrifuge-app/src/components/DataTable.tsx
jpangelle: Done!

centrifuge-app/src/components/Portfolio/usePortfolio.ts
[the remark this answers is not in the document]
jpangelle: https://github.com/centrifuge/apps/assets/28536523/e445f4ed-0e26-4de0-9782-5567d2118690

Yeah, unfortunately it is a bit janky when implementing it this way.

onnovisser (approved):

PR #1676, opened 2023-11-02
fix: use new window

### LINKED ISSUE #1664: Gettting stuck at KYC step when onboarding with Safe App
When onboarding via Safe, the app will not usher the user forward when completing a step that contains an `iframe` which is when the user does KYC or manual KYB. The page gets stuck here:
image

When a user is verified or declined, ShuftiPro uses `window.postMessage` to inform the parent window. Therefore, this [attached event listener](https://github.com/centrifuge/apps/blob/main/centrifuge-app/src/pages/Onboarding/KnowYourCustomer/index.tsx#L69-L76) may need a refactor so that the app can properly listen for the post inside the `iframe`. This is tricky because Safe runs the Centrifuge app inside an `iframe` and the Centrifuge app runs the ShuftiPro steps inside an `iframe`. So it is an `iframe` inside an `iframe`.

jpangelle:
### Description

This pull request refactors onboarding to open the global onboarding flow in a new window when using the app with [Safe](https://app.safe.global/).

Closes #1664 

### Approvals

- [x] Dev
- [x] Product

### Impact

Onboarding

### DIFF
--- a/centrifuge-app/src/components/OnboardingAuthProvider.tsx
+++ b/centrifuge-app/src/components/OnboardingAuthProvider.tsx
@@ -35,10 +35,10 @@ export function OnboardingAuthProvider({ children }: { children: React.ReactNode
   const { data: session, refetch: refetchSession } = useQuery(
     ['session', selectedAccount?.address, proxy?.delegator, selectedAddress],
     () => {
-      // if user comes from Safe
+      // if user comes from external app
       if (isExternal) {
-        const safeSignatureSession = sessionStorage.getItem('external-centrifuge-onboarding-auth')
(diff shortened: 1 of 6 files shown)


centrifuge-app/src/components/OnboardingAuthProvider.tsx
```diff
@@ -35,10 +35,10 @@ export function OnboardingAuthProvider({ children }: { children: React.ReactNode
   const { data: session, refetch: refetchSession } = useQuery(
     ['session', selectedAccount?.address, proxy?.delegator, selectedAddress],
```
onnovisser, line 36: `isExternal` needs to be in the array too

onnovisser (approved):

PR #1718, opened 2023-11-30
change subgraph url `prod` instead of versioned URL

AStox:
### Description

This pull request updates the subgraph URL back to the original https://graph.centrifuge.io/tinlake url, which is now redirecting to goldsky correctly.

### Approvals

- [ ] Dev


gpmayorga (changes requested):
Actually, it is not redirecting properly

Screenshot 2023-12-04 at 18 15 21

gpmayorga (approved):
Looking good now

Issue #1713, opened 2023-11-29
Change face value and settlement price input to 8 decimals


annamehr:
## Background
* Currently the settlement price input in the finance component support 5 decimal places after the comma 
* T-bill tradfi broker support 8 decimal places
* This is a featured required for Anemoy

![Image](https://github.com/centrifuge/apps/assets/31280003/f48e0174-b9ba-446b-b771-e16494a4ab30)

## Requirements
* In the finance form change the input from 5 (xx.yyyyy) to 8 decimals (xx.yyyyyyyy)
* Entering less than 8 decimals also works

PR #1744, opened 2023-12-08
Centrifuge App: Use 8 decimals for face value and settlement price

onnovisser:
### Description

Also replaces the inputs in the finance and repay forms with the new inputs, mainly because the new currency input is less buggy.

#1713 

### Approvals

- [ ] Dev

### Screenshots

image


sophialittlejohn (approved):
🥳

PR #1882, opened 2024-01-18
Fix token price chart

onnovisser:
### Description

data beyond december wasn't showing in the price charts, because of the subquery limit. expands the daily pool states query to also return the tranche states, as this query already fetches data beyond the 100 item limit. uses that data for the tranche token price chart

### Approvals

- [ ] Dev

### Screenshots

image


sophialittlejohn (commented):
These changes seem to have broken the tinlake pool pages (e.g https://app-pr1882.k-f.dev/pools/0x60b71e9DCEeDAAC275c377630E054bc60a21A02B) and looks like they still work on dev.
I'm seeing an error from the `useDailyPoolStates` hook:
```
Only works with Centrifuge Pools
```

onnovisser:
@sophialittlejohn Fixed

sophialittlejohn (approved):

Issue #1905, opened 2024-01-25
Display archived pools v1


annamehr:
[Figma file](https://www.figma.com/file/ng7qdNcSCXSDA6ZUdWIs6u/Pool-Overview%2F-Pool-Detail?type=design&node-id=1286-38476&mode=design&t=z9V0Kd9xhtY4mxeZ-0)

## Background
Pool Management requested the following change.

## Requirements
* Remove `Archived` filter from Pool status dropdown
* Add a link below the pool list `View archived pools`
* An accordion style list should open showing all archived pools

The following pools will be displayed behind the `View archived pools` link

- Branch Series 3
- databased.finance 1
- Fortunafi 1
- Fortunafi 2
- 1754 Factory Bling
- Cauris
- Consol Freight
- Flow Carbon
- Gig economy pool
- Harbor trade 2
- Reif pool
- BT1
- BT2

![Image](https://github.com/centrifuge/apps/assets/31280003/46d71ecb-673b-4047-9f99-5e4c785b8301)

TimHoub:
TBD which solution we'll take, but here is the design file. 

https://www.figma.com/file/ng7qdNcSCXSDA6ZUdWIs6u/Pool-Overview%2F-Pool-Detail?type=design&node-id=1289-35064&mode=design&t=fgPoY7l5SpLcjKqw-0

PR #2004, opened 2024-03-11
archived pools toggle

onnovisser:
### Description

Adds separate table with archived pools and a toggle to show/hide them

#1905

### Approvals

- [ ] Dev

### Screenshots

image


sophialittlejohn (commented):
Nice work! Only issue I see is that the table headers are still visible even if there are no archived pools

onnovisser:
@sophialittlejohn fixed

jpangelle (approved):

PR #2006, opened 2024-03-11
Convert address page

onnovisser:
### Description

Create a page (at `/convert`) with the address converter that's already in the debug panel. 

### Approvals

- [ ] Dev

### Screenshots

image


sophialittlejohn (approved):
Code looks good! do we need to link the new page somewhere?

onnovisser:
Didn't see any mention of a menu link to the page in the specs. I suppose it's just a url we can send users who need it 🤷

PR #2042, opened 2024-03-25
Fix pool proposal

onnovisser:
### Approvals

- [ ] Dev


centrifuge-js/src/modules/pools.ts
```diff
@@ -943,7 +943,7 @@ export function getPoolsModule(inst: Centrifuge) {
             if (options?.createType === 'propose') {
               const proposalTx = api.tx.utility.batchAll([
                 api.tx.preimage.notePreimage(tx.method.toHex()),
-                api.tx.democracy.propose({ Inline: tx.method.hash }, api.consts.democracy.minimumDeposit),
+                api.tx.democracy.propose({ Lookup: tx.method.hash }, api.consts.democracy.minimumDeposit),
```
wischli, line 946: When using `Lookup`, the preimage length is required. Thus, the current change would not work.

Don't know from the top of my head how to derive it from the `tx` here due to lacking Polkadot JS knowledge. I think the cleaner solution is to stick with `Inline` and just use `tx.toHex()` which then submits the entire call:
```suggestion
                api.tx.democracy.propose({ Inline: tx.toHex() }, api.consts.democracy.minimumDeposit),
```

wischli (changes requested):

centrifuge-js/src/modules/pools.ts
onnovisser: with `Inline` do we still need the `notePreimage` or can we only do the `propose`

mustermeiszer: But aren't the proposals already to long? I think the max was 128 bytes. 

you should be able to call `encode()` IIRC or use `(toHex().len()-2) / 2`

wischli: Couldn't find a getter to determine the length of the `hash` but since it is hexadecimal and the length bound in number of bytes, it should be `tx.toHex() / 2 - 1`. So an alternate solution would be

```suggestion
                api.tx.democracy.propose({ Lookup: { hash: tx.hash, len: tx.toHex().length / 2 - 1 }, api.consts.democracy.minimumDeposit),
```

onnovisser: yes the proposal is too long. the one for New Silver is 268 bytes. I see a property `tx.method.encodedLength` which looks to be the number of bytes

wischli: > with `Inline` do we still need the `notePreimage` or can we only do the `propose`?

Yes, that's the major selling point of `Inline`. @mustermeiszer is right though that the max bounded call length is 128 bytes. The current `register` call has 268 bytes.

onnovisser: updated it to include the length

centrifuge-js/src/modules/pools.ts
```diff
@@ -943,7 +943,10 @@ export function getPoolsModule(inst: Centrifuge) {
             if (options?.createType === 'propose') {
               const proposalTx = api.tx.utility.batchAll([
                 api.tx.preimage.notePreimage(tx.method.toHex()),
-                api.tx.democracy.propose({ Inline: tx.method.hash }, api.consts.democracy.minimumDeposit),
+                api.tx.democracy.propose(
+                  { Lookup: [tx.method.hash, tx.method.encodedLength] },
```
wischli, line 947: Did you test this call? I don't think it works as `Lookup` expects an object, i.e.
```suggestion
                  { Lookup: { hash: tx.method.hash, len: tx.method.encodedLength} },
```

onnovisser: I tested and a tuple seemed to work: https://polkadot.js.org/apps/?rpc=wss://fullnode.development.cntrfg.com#/explorer/query/0x23cba13ab8577309b22854fca0bacb04555a1ac4c2834543ffc3c1d527538f6c

wischli (approved):
Thanks for the testing reference, LGTM!

centrifuge-js/src/modules/pools.ts
wischli: Awesome, thanks!

Issue #2071, opened 2024-04-05
Bug interest payments


annamehr:
- [x] Fix interest payments 

https://github.com/centrifuge/apps/blob/main/centrifuge-app/src/pages/Loan/RepayForm.tsx#L66 this should use the portfolio RT API to get the outstanding principal + interest, then first repay the interest entirely including a small margin, and only then repay any principal

PR #2072, opened 2024-04-05
Repay form: pay interest first

onnovisser:
### Description

Pay interest (with a small margin) before paying the principal

#2071

### Approvals

- [ ] Dev


centrifuge-js/src/modules/pools.ts
```diff
@@ -3184,6 +3186,7 @@ export function getPoolsModule(inst: Centrifuge) {
               ? pricingInfo.valuationMethod.discountedCashFlow
               : undefined
           return {
+            fetchedAt: new Date(),
```
sophialittlejohn, line 3189: I don't understand how this would work, wouldn't it just use the current date every time the loans are fetched? Wouldn't that make the `time` you're calculating above always very small?

onnovisser: yes it could only be a few minutes worth, but it could also be hours if the tab stays open and no loan is updated, so loans aren't refetched. maybe the amount of interest accrued in the meantime is too small to notice, but maybe there's a fastidious issuer who'll wonder why not all interest was repaid. idk maybe it's overkill 🤷

sophialittlejohn: I like the idea! Maybe it would be smoother to refetch the loans before submitting the tx instead to make sure the interest is calculated properly?

onnovisser: With the subscriptions, there isn't a straightforward way to refetch, so this seemed like the most practical option.

sophialittlejohn: fair, if you could just leave a comment here explaining i to the next developer like you did to me that would be amazing! 🙏

onnovisser: added a comment

sophialittlejohn (approved):

Issue #1999, opened 2024-03-04
Mobile view pool detail page


annamehr:
[Figma design](https://www.figma.com/file/gIegw8r0OJk1kxSnZmX3zi/Mobile-design?type=design&node-id=1-81&mode=design&t=Dr1G87bKfWx6V7fk-0)

# Requirements
- [ ] remove the "get started" call to action on the pool list
- [ ] change pool table to only show name and value
- [ ] pool details: just stack cards
- [ ] remove the "footer"

![Image](https://github.com/centrifuge/apps/assets/31280003/13c922ff-cf66-41b1-b349-13be7c64a689)

PR #2085, opened 2024-04-15
chore: improve mobile view

jpangelle:
### Description

This pull request improves the mobile view of the home page and the pool detail page.

#1999

### Approvals

- [ ] Dev
- [ ] Designer
- [x] Product


centrifuge-app/src/components/LayoutBase/index.tsx
```diff
@@ -26,6 +27,8 @@ type LayoutBaseProps = {
 }
 
 export function LayoutBase({ children, gap }: LayoutBaseProps) {
+  const screenSize = useScreenSize()
```
onnovisser, line 30: there's already a hook that checks screen size that also takes theme values into account
`const isMedium = useIsAboveBreakpoint('M')`
M is 900px

TimHoub:
**Pool listing page:**
- We should use the same way of visualising the pools as we do on desktop -> not the standard table but the card-like table design
 - For the pool card-table we show 'pool name' & 'Value locked' on mobile
 
 **Menu**
 - bottom menu icons should have a pixel size of 24px + tapable area of 45px. (Added it in the screenshot with the orange square in bottom menu)
 
 
Screenshot 2024-04-15 at 11 03 36


**Pool detail page**

- Horizontal scroll on the entire page should be removed. Doesn't feel right at the moment. (view screenshot)
- Invest card doesn't render on mobile (please follow design where possible) 
- Add cashflow chart and transaction table back in the design. (for table I would activate horizontal scroll but only on the table for transactions if possible)
Screenshot 2024-04-15 at 11 15 52
IMG_1119

onnovisser (approved):

PR #2100, opened 2024-04-26
IPFS Compatibility

onnovisser:
### Description

Makes the App compatible with deployment to IPFS

### Approvals

- [ ] Dev


sophialittlejohn (approved):

onnovisser:
We really need the hash router, because loading the page from a subpath is broken now

PR #2144, opened 2024-05-21
Fix Transfer Debt

onnovisser:
### Description

Transferring from cash assets didn't work, because the order of operations for calculating principal was wrong

### Approvals

- [ ] Dev


sophialittlejohn (approved):
😅

Issue #2183, opened 2024-06-04
Write pool creation script for e2e testing


annamehr:
[Hackmd](https://centrifuge.hackmd.io/wgjyqOKwR46HurZ_MQCZjg)

This script should set up a pool in 1 click and complete the following actions.
* Fill out create pool form
* Add asset originator 
* Create asset template 
* Invest into the pool
* Mint asset
* Finance asset 
* Repay asset

PR #2215, opened 2024-06-13
1 click debug pool creation

onnovisser:
### Description

Create a pool with 1 click from the debug panel. also adds a button to fund the accounts that are doing the pool creation/borrowing/investing etc.

- [x] create pool
- [x] Add asset originator and borrower hot wallet
- [x] Create asset template
- [x] Invest into the pool
- [x] Mint asset
- [x] Add fees
- [x] Feed oracle values
- [x] Close epoch
- [x] Finance asset
- [x] Repay asset

#2183

### Approvals

- [ ] Dev


centrifuge-app/src/components/DebugFlags/components/CreateAndSetupPool.tsx
```diff
@@ -0,0 +1,627 @@
...
+            //   kycRestrictedCountries: ['us', 'um'],
+            //   kybRestrictedCountries: ['us'],
+            //   externalOnboardingUrl: 'https://www.anemoy.io/funds/ltf#get-access',
+            //   podReadAccess: true,
+            //   taxInfoRequired: true,
+            // },
+            loanTemplates: [{ id: TEMPLATE_HASH, createdAt: new Date().toISOString() }],
+          }
+          const metadataInput: Partial<PoolMetadataInput> = {
+            poolName: 'E2E Test Pool',
```
sophialittlejohn, line 194: Would be super helpful if we could somehow give each new pool a different name. What do you think about using a random name generator?

PR #2216, opened 2024-06-14
Asset detail page redesign

hieronx:
### Description

This pull request...

Closes #<issue-number>

### Approvals

- [ ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product


centrifuge-app/src/components/Charts/AssetPerformanceChart.tsx
```diff
@@ -0,0 +1,291 @@
...
+  const dataUrl: any = React.useMemo(() => {
+    if (!assetSnapshots || !assetSnapshots?.length) {
+      return undefined
+    }
+
+    const formatted = assetSnapshots.map((snapshot) => {
+      return {
+        ...snapshot,
+      }
+    })
```
sophialittlejohn, line 60: is this still necessary? looks like it's just mapping to the same array

sophialittlejohn (approved):
Love the future line! Super cool :) 

2 things I noticed:
- we don't have a chip/label that lets the user know at a glance what the status is of the asset, do we want to add it back somewhere?
- Anemoy [asset 1](https://pr2216-app-ff-production.k-f.dev/pools/4139607887/assets/1) doesn't show the value for most data points, while all other assets seems to be working fine

One improvement not necessarily for now since we have it in a few places: prevent the layout jump, i.e reserve the space for the chart while it loads. Adding a ticket for it!

centrifuge-app/src/components/Charts/AssetPerformanceChart.tsx
hieronx: Good catch, fixed!

onnovisser (approved):

Issue #2221, opened 2024-06-18
Use EVM address inputs for issuers


hieronx:
Use the address input that we have for CFG transfers, that supports both EVM addresses and Centrifuge addresses, for:
- [x] Pool managers on Create pool page
- [x] Investor input for Centrifuge Chain on Investors tab
- [x] AO wallets on Access tab
- [x] Receiving address on Fees tab

Side note: this was already done for the for input in the portfolio page, logic should be copyable from there: https://app.centrifuge.io/portfolio?send=CFG

Converting an EVM address to a substrate one should be done like in the receive tab https://app.centrifuge.io/portfolio?receive=CFG

0x4Graham:
I've tested this the best I could, working 👍

PR #2226, opened 2024-06-19
Use evm inputs for issuers

kattylucy:
Use the address input that we have for CFG transfers, that supports both EVM addresses and Centrifuge addresses

#2221 

### Approvals

- [ ] Dev
- [x ] Dev
- [ ] Designer
- [ x] Product

### Impact

Pool managers on Create pool page
Investor input for Centrifuge Chain on Investors tab
AO wallets on Access tab
Receiving address on Fees tab

### DIFF
--- a/centrifuge-app/src/pages/IssuerPool/Access/AssetOriginators.tsx
+++ b/centrifuge-app/src/pages/IssuerPool/Access/AssetOriginators.tsx
@@ -1,6 +1,7 @@
 import {
   addressToHex,
   computeTrancheId,
+  evmToSubstrateAddress,
   getCurrencyLocation,
   isSameAddress,
   PoolMetadata,
@@ -8,6 +9,7 @@ import {
   WithdrawAddress,
 } from '@centrifuge/centrifuge-js'
 import {
+  useCentEvmChainId,
   useCentrifuge,
   useCentrifugeApi,
   useCentrifugeConsts,
@@ -118,6 +120,7 @@ function AOForm({
   poolId: string
 }) {
   const [isEditing, setIsEditing] = React.useState(false)
+  const chainId = useCentEvmChainId()
   const [account] = useSuitableAccounts({ poolId, actingAddress: [ao.address] }).filter((a) => a.proxies?.length === 2)
   const identity = useIdentity(ao.address)
   const api = useCentrifugeApi()
@@ -311,8 +314,8 @@ function AOForm({
       }
       execute(
         [
-          addedWithdraw.map((w) => getKeyForReceiver(api, w)),
-          removedWithdraw.map((w) => getKeyForReceiver(api, w)),
+          addedWithdraw.filter((w) => Object.keys(w).length !== 0).map((w) => getKeyForReceiver(api, w)),
+          removedWithdraw.filter((w) => Object.keys(w).length !== 0).map((w) => getKeyForReceiver(api, w)),
           null,
           addedPermissions,
           addedDelegates,
@@ -438,7 +441,9 @@ function AOForm({
                       <AddAddressInput
                         existingAddresses={[...form.values.delegates, ao.address]}
                         onAdd={(address) => {
-                          fldArr.push(addressToHex(address))
+                          fldArr.push(
+                            isEvmAddress(address) ? evmToSubstrateAddress(address, chainId ?? 0) : addressToHex(address)
+                          )
                         }}
                       />
                     )}
--- a/centrifuge-app/src/pages/IssuerPool/Configuration/AddAddressInput.tsx
+++ b/centrifuge-app/src/pages/IssuerPool/Configuration/AddAddressInput.tsx
@@ -1,6 +1,6 @@
-import { isSameAddress } from '@centrifuge/centrifuge-js'
+import { addressToHex, isSameAddress } from '@centrifuge/centrifuge-js'
 import { useCentrifugeUtils } from '@centrifuge/centrifuge-react'
-import { Button, Grid, SearchInput, Shelf, Text } from '@centrifuge/fabric'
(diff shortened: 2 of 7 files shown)


centrifuge-app/src/pages/IssuerPool/Configuration/AddAddressInput.tsx
```diff
@@ -19,19 +19,27 @@ export function AddAddressInput({
...
 
+  function handleChange(e: React.ChangeEvent<HTMLInputElement>) {
+    const newAddress = e.target.value
+    setAddress(newAddress)
+  }
+
+  function handleBlur(e: React.FocusEvent<HTMLInputElement>) {
+    const address = e.target.value
+    if (address && !exists) {
+      onAdd(addressToHex(address))
```
onnovisser, line 35: Is this the desired behaviour? If so, it should check that it's a valid address (check if `truncated` is truthy)

fabric/src/components/TextInput/index.tsx
```diff
@@ -252,7 +252,12 @@ export function TextAreaInput({
 }
 
 export type AddressInputProps = Omit<React.InputHTMLAttributes<HTMLInputElement>, 'value'> &
-  Omit<InputUnitProps, 'inputElement'>
+  Omit<InputUnitProps, 'inputElement'> & {
+    value?: string
+    onChange: (e: React.ChangeEvent<HTMLInputElement>) => void
+    onBlur?: (e: React.FocusEvent<HTMLInputElement>) => void
```
sophialittlejohn, line 258: I don't think you have to explicitly declare these three types, they should already be included in the type `React.InputHTMLAttributes<HTMLInputElement`

sophialittlejohn (commented):
Great job! 

Can you please make sure the add address button is vertically centered? And the same for the pool managers?
Screenshot 2024-06-21 at 12 34 06

In the fees tab I noticed that when I use an EVM address it's saying "Invalid address", I think you might need to adjust the validation to make it work for EVM
Screenshot 2024-06-21 at 12 38 45

fabric/src/components/TextInput/index.tsx
kattylucy: Yes, you are right ! - thanks

sophialittlejohn:
Nice work Katty! Just a few more things to clarify and then I think it's good to go!

I think we can now remove the dropdown next to the input since the network is now implied by the address, buuuut @hieronx will we still need it to differentiate between the different EVM chains for liquidity pools?
Screenshot 2024-06-25 at 13 48 46

I think we also need some kind of icon for the evm wallets here, maybe just the ethereum logo for now? I also tried submitting the transaction but it failed: `{"err":{"cannotLookup":null}}`. Maybe you're still missing the address conversion here? 
Screenshot 2024-06-25 at 13 51 11

hieronx:
> I think we can now remove the dropdown next to the input since the network is now implied by the address, buuuut @hieronx will we still need it to differentiate between the different EVM chains for liquidity pools?

Yeah we still need to differentiate between EVM chains. For the Centrifuge network, the input could be either an EVM address or a Substrate/Centrifuge address. For all other networks, it's always an EVM address.

kattylucy:
_Yeah we still need to differentiate between EVM chains._

How do we currently differentiate the two?

hieronx:
@kattylucy that was in response to Sophia's question. The dropdown is for selecting the network

sophialittlejohn:
Almost there, typing doesn't work anymore on this input

Screenshot 2024-07-18 at 17 24 54

centrifuge-app/src/pages/IssuerPool/Access/AssetOriginators.tsx
```diff
@@ -156,6 +158,15 @@ function AOForm({
...
   } = useCentrifugeConsts()
 
+  React.useEffect(() => {
+    const fetchChainId = async () => {
+      const id = await cent.getChainId()
+      setChainId(id)
+    }
+
+    fetchChainId()
+  }, [cent])
```
sophialittlejohn, line 168: this would make a good hook

sophialittlejohn (approved):
Left one more comment to clean up a bit but already looks good to me 🔥 nice work pushing this one through!

centrifuge-app/src/pages/IssuerPool/Access/AssetOriginators.tsx
```diff
@@ -156,6 +158,15 @@ function AOForm({
     transferAllowlist: { receiverDeposit },
   } = useCentrifugeConsts()
 
+  React.useEffect(() => {
+    const fetchChainId = async () => {
+      const id = await cent.getChainId()
```
onnovisser, line 163: this is the wrong chain id. you'll need to use `useCentEvmChainId()`

fabric/src/components/TextInput/index.tsx
```diff
@@ -262,40 +265,47 @@ export function AddressInput({
...
   onBlur,
   onChange,
+  value = '',
+  clearIcon,
   ...inputProps
 }: AddressInputProps) {
   const defaultId = React.useId()
   id ??= defaultId
 
+  const [inputVal, setInputVal] = React.useState<string>(value)
```
onnovisser, line 275: the state doesn't seem necessary. the effect can use the `value` prop. using state here can lead to a bug when the value is changed outside the component, causing the effect not to run

onnovisser (approved):

centrifuge-js/src/utils/index.ts
```diff
@@ -137,6 +137,8 @@ export function computeMultisig(multisig: Multisig): ComputedMultisig {
 }
 
 export function evmToSubstrateAddress(address: string, chainId: number) {
+  if (!chainId) return
```
onnovisser, line 140: better to throw in that case

Issue #2232, opened 2024-06-20
Exclude cash assets from asset counts


hieronx:

kattylucy:
Hi @hieronx - do you mind adding a bit of context here? do we count cash assets in the total assets? I see this is not true for all the pools (right ?. How can I set up a test for this? TY @hieronx

hieronx:
<img width="1533" alt="Screenshot 2024-07-01 at 15 35 41" src="https://github.com/centrifuge/apps/assets/1748621/5bc2ecf3-ac08-4b66-9c67-5aa9ce36384a">

If you look at the Anemoy assets now. It is showing 10 total assets and 6 ongoing assets. This is including the `Settlement Account`, which is a cash asset. It should exclude this, so the total assets should be 9 and the ongoing assets should be 5.

kattylucy:
Is settlement account the only type of cash asset? @hieronx

kattylucy:
Ah perfect. Thanks @hieronx

PR #2259, opened 2024-07-01
Remove cash assets from ongoing & total assets count

kattylucy:
Total assets and Ongoing assets count must exclude cash assets

#2232 

- [ x ] Dev
- [ ] Dev
- [ ] Designer
- [ x ] Product

- Check the overview pool key metrics assets should exclude cash assets for total & ongoing assets
- Check pool -> assets tab -> key metrics should exclude cash assets for both total and ongoing


hieronx (approved):
LGTM

Issue #1838, opened 2023-12-27
Download links for wallets should be based on users browser and not hard coded to chrome


sophialittlejohn:
https://centrifugedao.slack.com/archives/C04HCJRJUSY/p1703597961845459

0x4Graham:
Looks good. LGTM

pravindev12:
@sophialittlejohn @kattylucy this bug is closed or still active?

PR #2265, opened 2024-07-04
Links for wallets

kattylucy:
- Show non-support banner regardless if user wallet is connected. Do not show the banner if browser is chrome/firefox
- Links to firefox store shall be different than the url for chrome store. Unfortunately '@subwallet/wallet-connect/types' only returns walletUrl for chrome only, some hard code links was required to make it work on firefox.

#1838 

- [ x ] Dev
- [ ] Dev
- [ ] Designer
- [ ] Product

### DIFF
--- a/centrifuge-react/src/components/WalletProvider/WalletDialog.tsx
+++ b/centrifuge-react/src/components/WalletProvider/WalletDialog.tsx
@@ -1,3 +1,4 @@
+import { getSupportedBrowser } from '@centrifuge/centrifuge-app/src/utils/getSupportedBrowser'
 import {
   Button,
   Card,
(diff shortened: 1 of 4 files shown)


centrifuge-react/src/components/WalletProvider/evm/connectors.ts
```diff
@@ -128,6 +129,7 @@ export function getEvmConnectors(
       get shown() {
         return !isMobile() || this.installed
       },
+      extensionName: 'metamask'
```
onnovisser, line 132: why not use the existing `id` for this?

onnovisser: mainly saying it because the typing isn't quite right now, as it's casting to `EvmConnectorMeta`, but not all connectors have the `extensionName`. I'd prefer using `id` and then doing a check for it in `getAdjustedInstallUrl`

centrifuge-react/src/components/WalletProvider/evm/connectors.ts
```diff
@@ -26,6 +26,7 @@ export type EvmConnectorMeta = {
   id: string
   title: string
   installUrl: string
+  extensionName: string
```
onnovisser, line 29: leftover extensionName

centrifuge-react/src/components/WalletProvider/WalletDialog.tsx
```diff
@@ -40,6 +41,24 @@ const title = {
...
+
+const getAdjustedInstallUrl = (wallet: { extensionName: string; installUrl: string }): string => {
+  const browser = getSupportedBrowser()
+  const { extensionName, installUrl } = wallet
+  if (browser === 'firefox') {
+    return `https://addons.mozilla.org/en-US/firefox/addon/${walletsList[extensionName || 'metamask']}/`
+  } else {
+    return installUrl
+  }
+}
```
onnovisser, line 60:
```suggestion
function getAdjustedInstallUrl(wallet: Wallet | EvmConnectorMeta): string {
  const browser = getSupportedBrowser()
  if (browser === 'firefox') {
    return `https://addons.mozilla.org/en-US/firefox/addon/${walletsList['extensionName' in wallet ? wallet.extensionName : 'metamask']}/`
  } else {
    return wallet.installUrl
  }
}
```

onnovisser (approved):

Issue #2263, opened 2024-07-03
Asset field fixes


hieronx:
Looking at this asset: https://app-demo.k-f.dev/issuer/3552744361/assets/1

- [ ] `Fund Short Description` should be fully shown, taking up multiple rows
- [ ] `Fund Link` should automatically become `<a>` with `target="_blank"`

pravindev12:
@kattylucy is this issue fixed, by going through the comments i think it's fixed and merged. i am planning to contribute.

PR #2266, opened 2024-07-04
Add fixes to assets fields table

kattylucy:
- Fund Short Description should be fully shown, taking up multiple rows
- Fund Link should automatically become <a> with target="_blank"

#2263 

- [ x ] Dev
- [ ] Dev
- [ ] Designer
- [ x ] Product

Screenshot 2024-07-04 at 10 48 09 AM

### DIFF
--- a/centrifuge-app/src/pages/Loan/index.tsx
+++ b/centrifuge-app/src/pages/Loan/index.tsx
@@ -250,7 +250,10 @@ function Loan() {
                               const attribute = templateData.attributes?.[key]!
                               const value = publicData[key]
                               const formatted = value ? formatNftAttribute(value, attribute) : '-'
-                              return { label: attribute.label, value: formatted }
+                              return {
+                                label: attribute.label,
+                                value: formatted,
+                              }
                             })}
                         />
                       </Stack>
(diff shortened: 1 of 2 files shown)


centrifuge-app/src/pages/Loan/index.tsx
```diff
@@ -250,7 +250,12 @@ function Loan() {
                               const attribute = templateData.attributes?.[key]!
                               const value = publicData[key]
                               const formatted = value ? formatNftAttribute(value, attribute) : '-'
-                              return { label: attribute.label, value: formatted }
+                              return {
+                                label: attribute.label,
+                                value: formatted,
+                                multirow: attribute.label.includes('Description'),
```
hieronx, line 256: We need to make this more generic. I would just detect length of the string and use that to determine `multirow`.

And use an URL regex for `asLink` (regex doesn't need to be perfect, as long as it catches simple URLs)

kattylucy: Ah good idea.

hieronx (commented):
From product perspective, looks good! 🔥

onnovisser (approved):

Issue #2007, opened 2024-03-11
Upgrade react-router


sophialittlejohn:
react-router: allow for easier query params (for reports)

Use hash router - so that routing works with IPFS deployment (Katty will check with Onno)

kattylucy:
I think we need to separate this ticket into two tickets to make sure we don't break anything because react-router upgrade also have a lot of work changes @sophialittlejohn

PR #2268, opened 2024-07-05
React router v6 update

kattylucy:
- Upgrade react-router-dom

#2007 

- [ x ] Dev
- [ x ] Dev
- [ ] Designer
- [ ] Product

The changes here impact the app entirely, if we have a component using react-route or react-router-dom it will be affecting. I did my best to test the app but would be nice to have pair of extra eyes.

NOTE: when switching branches you might encounter issues with the cache of node-modules. do run the following command to fix the issue: 

1) rm -rf node_modules
2) yarn install
3) yarn start


kattylucy:
This one is a beast 👹 - Let's take the time to test the app. I added hash as well, thank to Onno for the context on IPFS!

centrifuge-app/src/components/DataTable.tsx
```diff
@@ -141,7 +141,7 @@ export const DataTable = <T extends Record<string, any>>({
         <DataRow
           hoverable={hoverable}
           as={onRowClicked ? Link : 'div'}
-          to={onRowClicked && (() => onRowClicked(row))}
+          to={onRowClicked ? onRowClicked(row) : undefined}
```
onnovisser, line 144: can you add a TODO to rename this prop? We should rename to something like `getRowLink` or something. The current name suggests that the callback is only called when the row is clicked

centrifuge-react/package.json
```diff
@@ -69,7 +69,9 @@
     "bn.js": "^5.2.1",
     "buffer": "^6.0.3",
     "decimal.js-light": "^2.5.1",
+    "history": "latest",
     "react-query": "^3.39.1",
+    "react-router-dom": "latest",
```
onnovisser, line 74: react-router and history packages shouldn't be part of centrifuge-react. the website also uses centrifuge-react

centrifuge-app/src/components/LoanList.tsx
```diff
@@ -55,9 +56,9 @@ const getLoanStatus = (loan: Loan | TinlakeLoan) => {
 
 export function LoanList({ loans }: Props) {
   const { pid: poolId } = useParams<{ pid: string }>()
-  const pool = usePool(poolId)
-  const isTinlakePool = poolId.startsWith('0x')
-  const basePath = useRouteMatch(['/pools', '/issuer'])?.path || ''
+  const pool = usePool(`${poolId}`)
```
onnovisser, line 59: Better to throw if `poolId` is `undefined` is such cases, to narrow the type of `poolId` to `string`, instead of `string | undefined`

centrifuge-app/src/utils/useLoans.ts
```diff
@@ -14,12 +14,12 @@ export function useLoans(poolId: string) {
   return isTinlakePool ? tinlakeLoans : centLoans
 }
 
-export function useLoan(poolId: string, assetId: string) {
-  const loans = useLoans(poolId)
-
-  return loans && [...loans].find((loan) => loan.id === assetId)
+export function useLoan(poolId: string | undefined, assetId: string | undefined) {
```
onnovisser, line 17: same here

centrifuge-app/src/utils/usePools.ts
```diff
@@ -18,14 +18,14 @@ export function usePools(suspense = true) {
 }
 
 export function usePool<T extends boolean = true>(
-  id: string,
+  id: string | undefined,
```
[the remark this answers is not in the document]
kattylucy: You mean throwing as in error? @onnovisser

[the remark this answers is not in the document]
onnovisser: I mean throwing in the place where the poolId is passed if the route param is not set. e.g.:
```
const { pid: poolId } = useParams<{ pid: string }>()
if (!poolId) throw new Error('Pool not found')
```
this way, poolId can't possibly be undefined when it's passed to usePool

package.json
```diff
@@ -31,6 +31,8 @@
     "@sendgrid/mail": "^7.7.0",
     "bn.js": "^5.2.1",
     "dotenv": "16.0.3",
+    "react-router": "6",
+    "react-router-dom": "6",
```
onnovisser, line 35: this shouldn't be needed

centrifuge-app/src/components/NavigationTabs.tsx
```diff
@@ -1,27 +1,28 @@
-import { Tabs, TabsItem, TabsItemProps } from '@centrifuge/fabric'
-import * as React from 'react'
-import { useRouteMatch } from 'react-router'
-import { Link } from 'react-router-dom'
+import { Tabs, TabsItem, TabsItemProps } from '@centrifuge/fabric';
```
onnovisser, line 1: you might need to set your editor to lint on save, so that Prettier runs

centrifuge-app/src/components/PoolOverview/TransactionHistory.tsx
```diff
@@ -49,15 +49,15 @@ export const columns = [
       return fromAssetId && toAssetId && activeAssetId == fromAssetId.split('-')[1] ? (
         <Text as="span" variant="body3">
           {fromAssetName} &rarr;{' '}
-          <AnchorTextLink target="_self" href={`/pools/${toAssetId?.split('-')[0]}/assets/${toAssetId?.split('-')[1]}`}>
+          <AnchorTextLink target="_self" href={`#/pools/${toAssetId?.split('-')[0]}/assets/${toAssetId?.split('-')[1]}`}>
```
onnovisser, line 52: this should be a RouterTextLink

centrifuge-app/src/utils/useBasePath.tsx
```diff
@@ -0,0 +1,16 @@
+import { useMatch } from 'react-router-dom';
```
onnovisser, line 1: looking great, just a couple more places where Prettier needs to run

kattylucy: Should we add a script to run prettier across files? @onnovisser

Issue #2223, opened 2024-06-18
Prevent Content Layout Shift


sophialittlejohn:
Reserve the space for charts while they're loading (maybe using aspect-ratio?)

- Chart on asset detail page
- Chart on pool detail page (pool performance)
- Liquidity tab on pool detail page (Originations & repayments, Investments & redemptions)
- Centrifuge prime detail page (currently graph is not loading due to undefined error, please fix this as well 😄)

PR #2279, opened 2024-07-11
Fix layout on assets detail page

kattylucy:
- Asset detail page:  Add < no data available > if data is empty to avoid layout emptiness/switching **(picture 1)**
- Liquidity page: Prevent layout switching by reorganizing rendering of component **(picture 2)**
- Prime page: Add box around graph layout to maintain dimensions. **(picture 3)**

#2223 

- [ x ] Dev

**Picture 1**
Screenshot 2024-07-11 at 11 24 02 AM

**Picture 2**

https://github.com/centrifuge/apps/assets/51223655/e0334def-5766-4eac-ba7f-43e4d07faafc

**Picture 3**

### DIFF
--- a/centrifuge-app/src/components/Charts/AssetPerformanceChart.tsx
+++ b/centrifuge-app/src/components/Charts/AssetPerformanceChart.tsx
@@ -139,13 +139,6 @@ function AssetPerformanceChart({ pool, poolId, loanId }: Props) {
 
   const isChartEmpty = React.useMemo(() => !data.length || assetSnapshots?.length < 1, [data, assetSnapshots])
 
-  const aggregatedData = data.reduce((acc, cur, index) => {
-    if (index % 2 === 0) {
-      acc.push(cur)
-    }
-    return acc
-  }, [])
-
   if (!assetSnapshots) return <Spinner style={{ margin: 'auto', height: 350 }} />
 
   return (
@@ -198,9 +191,9 @@ function AssetPerformanceChart({ pool, poolId, loanId }: Props) {
         )}
 
         <Shelf gap={4} width="100%" color="textSecondary">
-          {aggregatedData?.length ? (
+          {data?.length ? (
             <ResponsiveContainer width="100%" height={200} minHeight={200} maxHeight={200}>
-              <LineChart data={aggregatedData} margin={{ left: -36 }}>
+              <LineChart data={data} margin={{ left: -36 }}>
                 <defs>
                   <linearGradient id="colorPoolValue" x1="0" y1="0" x2="0" y2="1">
                     <stop offset="5%" stopColor={chartColor} stopOpacity={0.4} />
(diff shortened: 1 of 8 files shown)


sophialittlejohn:
This looks good! One more thing that would be great in this ticket is if we could prevent the vertical jumping of the page on load which happens while the charts are loading. There's an example [here](https://web.dev/articles/cls) that shows what I mean. Also I got the name wrong, the ticket was supposed to say fix `Cumulative layout shift (CLS)` instead of content, sorry! So the idea would be to reserve the space in advance for the charts so when they load they fill the space and don't cause the content beneath it to jump.

kattylucy:
Ok, so I did some research into CLS - The best way to fix it, as suggested is to add a fix height the container around the graph. Which I did, now, if the data is non-available for the graph we then have a big container with no data on it, I kept the changes and you can check asset performance graph to see it working. 

If we decide we don't want an empty container when no data is available and we want to hide the container (as it was prev), we need to be ok with the content switching in the x axis, so in this example, the key metrics card switching to the left. 

Let me know which option you prefer and I'll make the changes. Im personally fine with the fixed height to avoid any jumping either on the x or y axis of the page.

@sophialittlejohn

sophialittlejohn:
I think it's better to fix the height because we will generally assume that the subquery is working and the data will load. The table on the asset detail page looks good now when it loads, no more content jumping 🙌  if you could just do the same for the rest of the graphs that would be amazing!
- graph on pool detail page
- graph on prime page
- graph on portfolio page

kattylucy:
Yes, will do that - I was waiting to confirm which approach to take. Thank YOU! @sophialittlejohn

centrifuge-app/src/components/Charts/AssetPerformanceChart.tsx
```diff
@@ -137,29 +137,41 @@ function AssetPerformanceChart({ pool, poolId, loanId }: Props) {
...
-  if (!assetSnapshots) return <Spinner style={{ margin: 'auto' }} />
-  if (assetSnapshots?.length < 1) return null
+  const isChartEmpty = React.useMemo(() => !data.length || assetSnapshots?.length < 1, [data, assetSnapshots])
+
+  const aggregatedData = data.reduce((acc, cur, index) => {
+    if (index % 2 === 0) {
+      acc.push(cur)
+    }
+    return acc
+  }, [])
```
sophialittlejohn, line 147: Why is every second item being skipped in the data array?

kattylucy: I think it was mostly for the graphs with a lot of data to avoid showing every item and have more space for the dates. I changed it back 

Screenshot 2024-07-23 at 8 30 39 AM

sophialittlejohn: I think that should be handled by the charts lib (recharts) itself We've done it in some other charts. You can programmatically determine how many data points should be shown

kattylucy: Do we want to do that or do we want to keep the data as in the picture (with -35 degree) so it doesn't look all grouped together (impossible to read)? I'm open to both options but I'm not sure if we want to display all of the data anyways.

sophialittlejohn (approved):
Looks good!

Issue #2273, opened 2024-07-10
Asset list report update


hieronx:
### Data source
- [x] Move to using the subquery as source of truth, using `AssetSnapshot`s rather than the onchain data
- [x] Add a date picker at the top for the snapshots
- [x] Exclude cash assets

### Columns (public credit pools)
- Name (`asset.name`)
- Market value (`presentValue`)
- Face value (`notional * outstandingQuantity`)
- Quantity (`outstandingQuantity`)
- Market price (`currentPrice`)
- Maturity date (`actualMaturityDate`)
- Unrealized profit (`unrealizedProfitAtMarketPrice`)
- Realized profit (`sumRealizedProfitFifo`)

### Columns (private credit pools)
- Name (`asset.name`)
- Value (`presentValue`)
- Principal outstanding (`outstandingPrincipal`)
- Interest outstanding (`outstandingInterest`)
- Principal repaid (`totalRepaidPrincipal`)
- Interest repaid (`totalRepaidInterest`)
- Additional repaid (`totalRepaidUnscheduled`)
- Origination date (`actualOriginationDate`)
- Maturity date (`actualMaturityDate`)
- Valuation method (`asset.valuationMethod`)
- Advance rate (`asset.advanceRate`)
- Collateral value (`asset.collateralValue`)
- Probability of default (PD) (`asset.probabilityOfDefault`)
- Loss given default (LGD) (`asset.lossGivenDefault`)
- Discount rate (`asset.discountRate`)

kattylucy:
Is the pool credit type always required? I see some pools (on demo) with neither public or private. If this case is possible in some instances, which one should be default? @hieronx

hieronx:
@kattylucy that seems to be an issue with demo pools then, it should always be set to either public or private credit. But if it is not set, I would default to private credit.

kattylucy:
Great. Another question, should we also add a group by field to display a default amount of assets based on this field?

![Image](https://github.com/user-attachments/assets/c5ca8a21-8263-462d-bf0d-74a8ab0f6a7d)

@hieronx

hieronx:
I'm not sure I follow. What would the options in the Group by dropdown be for the asset list report?

PR #2289, opened 2024-07-18
Asset list report update

kattylucy:
Update asset list report 

- Use subquery instead of onchain for the snapshots
- Add new columns/value pairs
- Add new date filter for asset report 
- Set <ongoing> as default filter value

#2273 

- [ ] Dev

### DIFF
--- a/centrifuge-app/src/components/Report/AssetList.tsx
+++ b/centrifuge-app/src/components/Report/AssetList.tsx
@@ -20,8 +20,8 @@ const valuationLabels = {
   oracle: 'Fungible asset - external pricing',
 }
 
-function getColumnConfig(poolCreditType: string, symbol: string) {
-  if (poolCreditType === 'privateCredit') {
+function getColumnConfig(isPrivate: boolean, symbol: string) {
+  if (isPrivate) {
     return [
       { header: 'Name', align: 'left', csvOnly: false, formatter: noop },
       {
@@ -177,7 +177,11 @@ export function AssetList({ pool }: { pool: Pool }) {
   const { symbol } = pool.currency
 
   const snapshots = useAllPoolAssetSnapshots(pool.id, startDate)
-  const columnConfig = useMemo(() => getColumnConfig(poolCreditType, symbol), [poolCreditType, symbol])
+  const isPrivate = useMemo(
+    () => poolCreditType === 'Private credit' || poolCreditType === 'privateCredit',
+    [poolCreditType]
+  )
+  const columnConfig = useMemo(() => getColumnConfig(isPrivate, symbol), [poolCreditType, symbol])
 
   const columns = useMemo(
     () =>
@@ -230,7 +234,7 @@ export function AssetList({ pool }: { pool: Pool }) {
         return dateB - dateA
       })
       .map((snapshot) => {
-        if (poolCreditType === 'Private credit') {
+        if (isPrivate) {
           return {
             name: '',
             value: [
@@ -265,14 +269,13 @@ export function AssetList({ pool }: { pool: Pool }) {
               snapshot?.actualMaturityDate,
               snapshot?.unrealizedProfitAtMarketPrice,
               snapshot?.sumRealizedProfitFifo,
-              snapshot?.assetId,
             ],
             heading: false,
             id: snapshot?.assetId,
           }
         }
       })
-  }, [snapshots, poolCreditType, symbol, loanStatus])
+  }, [snapshots, isPrivate, symbol, loanStatus])
 
   useEffect(() => {
     if (!snapshots?.length) {
--- a/centrifuge-app/src/components/Report/AssetList.tsx
+++ b/centrifuge-app/src/components/Report/AssetList.tsx
@@ -117,7 +117,7 @@ function getColumnConfig(poolCreditType: string, symbol: string) {
         align: 'right',
         csvOnly: false,
         sortable: true,
-        formatter: (v: any) => formatBalance(v, symbol, 2),
+        formatter: (v: any) => formatBalance(v, '', 2),
       },
       {
         header: 'Market price',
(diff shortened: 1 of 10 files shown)


hieronx:
@kattylucy I was curious to test the new asset list report for the LTF pool (https://pr2289-app-ff-production.k-f.dev/#/pools/4139607887/reporting/asset-list) but ran into a few small issues, fixed those here: https://github.com/centrifuge/apps/pull/2323

Besides that, it looks great though! Nice job refactoring the centrifuge.js code to use the subquery data :)

One main comment: it would be great if the `Day` dropdown could default to the date of today. (but remain `Jan 1` of the year for all other reports).

kattylucy:
Awesome, thanks for the help - I will fix the date @hieronx

hieronx:
Seeing it now in action, a few more small comments:
- Let's remove the ID column, it doesn't add any useful information
- Would be great to make the Name column values clickable, going to the asset detail page (https://pr2289-app-ff-production.k-f.dev/#/pools/4139607887/assets/1)
- Would be great to, by default, sort by maturity date such that the furthest date in the future is at the top. I tried this already but it doesn't seem to work correctly.

And two issues I noticed in the DYF pool (see https://pr2289-app-ff-production.k-f.dev/#/pools/1655476167/reporting/asset-list, put date to `27/7/24`):
- If the `maturity date` is `undefined`, it should state `Open-end` rather than `Jan 1, 1970`. 
- If the `notional` is `undefined`, let's show `-` in the Face value column rather than 0.00 USDC.

sophialittlejohn:
Nice work! Also just tested it and noticed a few things:
1. the filters for active, repaid and overdue don't seem to work yet
2. some pools don't have data in the asset-list report despite having assets (e.g https://app-pr2289.k-f.dev/#/pools/3783931117/assets). The loans are only in "Created" state so I'm not sure if they should appear in the report or not?

kattylucy:
@sophialittlejohn 

The first one is definitely a bug, thanks for pointing out. The second question, I kept the previous logic regarding which loans to display, based on prev log we shouldn't display 'created' status loans. Do we want to change this? @hieronx @sophialittlejohn

hieronx:
@kattylucy I would include `Created` assets if you select Status `All`, but make the default Status filter `Ongoing`.

hieronx:
Ah ok, then just leave them out. Would still default the status to `Ongoing` so repaid assets don't show up by default.

hieronx:
@kattylucy looking great!

One more comment: could you change the quantity col formatting from `10,170.00 USDC` to `10,170.00`?

kattylucy:
@hieronx Oh no! Let me look into it! Thanks!!

kattylucy:
One thing here, when clicking like this 

reporting -> asset list report -> asset name 

we go to the route for the specific asset. When clicking back, it takes me to the asset list instead of the table for reporting, should we change the behavior so the back button takes back to the prev page based on the user last page?

hope my question makes sense 

@hieronx

hieronx:
> we go to the route for the specific asset. When clicking back, it takes me to the asset list instead of the table for reporting, should we change the behavior so the back button takes back to the prev page based on the user last page?

Good point. Seems to make sense to me to change it like you suggest. @sophialittlejohn any thoughts?

kattylucy:
@hieronx do you mind sending me the pool and the date you are using so I can confirm the issues? having a hard time reproducing the errors - thank you!

hieronx:
The face value I tested on https://pr2289-app-ff-production.k-f.dev/#/pools/1655476167/reporting/asset-list, with today's date.

The maturity dates I tested on https://pr2289-app-ff-production.k-f.dev/#/pools/4139607887/reporting/asset-list, also with today's date.

hieronx (approved):
Looks great to me!

centrifuge-js/src/types/subquery.ts
```diff
@@ -144,22 +144,36 @@ export type SubqueryAssetTransaction = {
 export type SubqueryAssetSnapshot = {
   __typename?: 'AssetSnapshot'
   asset: {
+    actualOriginationDate: number
```
onnovisser, line 147: The Subquery never returns Rates or CurrencyBalances so this type is now misleading.  `getAssetSnapshots` also uses this type and doesn't return any of the new additions, so consumers of that function may now expect values to be there that aren't there. I would suggest creating separate SubqueryPoolAssetSnapshot and PoolAssetSnapshot types to have the functions accurately reflect what they return

centrifuge-app/src/components/Report/AssetList.tsx
```diff
@@ -16,155 +18,267 @@ const valuationLabels = {
...
+            id: snapshot?.assetId,
+          }
+        }
+      })
+  }, [snapshots, isPrivate, symbol, loanStatus])
 
-  React.useEffect(() => {
-    if (!data.length) {
+  useEffect(() => {
+    if (!snapshots?.length) {
```
onnovisser, line 281: why check for `snapshots` now and not `data` ? data may end up having no items because of the snapshots being filtered

centrifuge-app/src/components/Report/AssetList.tsx
```diff
@@ -16,155 +18,267 @@ const valuationLabels = {
...
+  }
 }
 
 export function AssetList({ pool }: { pool: Pool }) {
-  const loans = useLoans(pool.id) as Loan[]
-  const { setCsvData, loanStatus } = React.useContext(ReportContext)
+  const basePath = useBasePath()
+  const { loanStatus, startDate, setCsvData } = useContext(ReportContext)
+  const { data: poolMetadata } = usePoolMetadata(pool)
+  const poolCreditType = useMemo(() => poolMetadata?.pool?.asset.class || 'privateCredit', [poolMetadata])
```
onnovisser, line 176: I would only keep the `useMemo` for `data`, the others here do more harm than good

onnovisser (approved):

Issue #2290, opened 2024-07-18
Remove Tokens tab from Pools page


sonam-jo:
Remove the Tokens tab from the Pools overview page. 

![Image](https://github.com/user-attachments/assets/1b177337-40fd-48df-9ee3-05007168ca60)

sonam-jo:
Yes let's remove Pools as well. 

For the Header, let's change the text to: 

> Pools on Centrifuge let investors earn yield from real-world assets

 given the box next to the chart will be removed too.

PR #2298, opened 2024-07-19
Remove tokens page and update title

kattylucy:
Remove tokens page and update title for pools

#2290 

- [ ] Dev

Screenshot 2024-07-19 at 9 04 20 AM


centrifuge-app/src/components/PoolsTokensShared.tsx
```diff
@@ -1,56 +1,32 @@
...
-          </Text>
-        </Stack>
-
-        <LoadBoundary>
-          <CardTotalValueLocked />
-        </LoadBoundary>
-
-        {isMedium && (
-          <Stack alignItems="end">
-            <MenuSwitch links={links} />
```
sophialittlejohn, line 49: You should be able delete the MenuSwitch component as well. Would be great if you could also remove the route and all the components that we're used on the tokens page. I'd also recommend getting rid of this file `PoolsTokensShared` since there is no more shared overview. A small refactor should do the job

onnovisser (approved):

sonam-jo:
@kattylucy the spacing for the pools table seems off - it stretches all the way across to the page. The columns should retain their original width from either side of the page. 

Screenshot 2024-07-22 at 16 43 58

Issue #2234, opened 2024-06-20
Allow 1/n multisigs


onnovisser:
Substrate multisigs have a minimum threshold of 2, so having multiple pool managers currently isn't possible.
Alter the pool managers form to use proxy delegates when the threshold is 1 and multisigs when it's higher.

0x4Graham:
Works for me. @hieronx mind having a quick QA to make sure please?

PR #2307, opened 2024-07-23
Allow multiple pool managers without a multisig

onnovisser:
### Description

Basically having a multisig with a threshold of 1. In reality it's not a multisig, but just having multiple addresses as the Any proxy on the pool admin proxy. 

[#2234](https://github.com/centrifuge/apps/issues/2234)

### Approvals

- [ ] Dev


sophialittlejohn (approved):
Didn't test it but code looks good! 🔥

Issue #2325, opened 2024-07-29
New asset mint: Dates end up with T+1 day entered


0x4Graham:
minted a new asset and the settlement date and the maturity date seem to be T-1 the date I entered

![image](https://github.com/user-attachments/assets/df7ee197-4303-4e3a-8f3b-ad7a10842cdc)

kattylucy:
Do you mind telling me which pool is this? @0x4Graham

kattylucy:
@0x4Graham In the screenshot, can you help me to figure out which one is settlement date?

kattylucy:
I figured! - Thank you!, should be ready for review @0x4Graham

0x4Graham:
Works for me

PR #2334, opened 2024-08-01
Date timezone causing error - convert to UTC

kattylucy:
Settlement date & maturity date format date function causing error when transforming dates from timestamp.

#2325 

- [ ] Dev


sophialittlejohn (approved):
🔥

Issue #2332, opened 2024-07-30
Remove maturity option from cash accounts


hieronx:
When selecting the `Cash` asset type on the `Create asset` page, the maturity date should be fixed to `Open-end` and not be shown as an input.

kattylucy:
Does this means we want to disabled the two other options? 

![Image](https://github.com/user-attachments/assets/86656b72-541c-4c47-81aa-760d5ad52b2e)

hieronx:
@kattylucy we should just remove the `Pricing` section including the maturity dropdown entirely.

0x4Graham:
@kattylucy on the initial request works as expected,

Based on @hieronx comment, it seems that the pricing section still displaying when cash asset selected: 



![Image](https://github.com/user-attachments/assets/39548937-2e98-4624-a756-790e65e33682)

kattylucy:
@0x4Graham Hmm something is not right, Im on develop (local env) and it seems to be working, which link did you check?

0x4Graham:
https://app-demo.k-f.dev/#/issuer/3521690895/assets/create

0x4Graham:
Ah, I see on Dev it's working 👍 Sorry!

PR #2338, opened 2024-08-02
Remove maturity options for cash assets

kattylucy:
### Description

If cash asset selected then maturity date should be open-end <always>

#2332 

### Approvals

- [ ] Dev


sophialittlejohn (approved):
👍

Issue #2330, opened 2024-07-30
DYF Pool, Asset name and Maturity Date not displaying


0x4Graham:
See here: https://kflabs.slack.com/archives/C063K04K1F0/p1722348541304579?thread_ts=1721742041.497799&cid=C063K04K1F0

![Image](https://github.com/user-attachments/assets/8d8eaca1-1e27-4c8e-b97e-5e61bfe02ca1)

 fix the asset names and maturity date on the NAV management tab, for the DYF pool.

kattylucy:
Looking at this, the date is null for those fields and that's why is returning '' - I fixed the <invalid date> to also show the empty string if null. 

The name of the asset seems to be purposely declared as the id. Do we want to change it anyways? @0x4Graham

sonam-jo:
I see the asset names are fixed on the app already. Don't think we need the Maturity date column for this table

kattylucy:
As per our conversation this ticket is now just about removing the maturity date for the table

![Image](https://github.com/user-attachments/assets/3a3cd82c-8d7f-48c9-8f07-fd8252991ed5)

0x4Graham:
Now that I think about it, we should most probably remove the Maturity date from the assets table in here as well https://app-ff-production.k-f.dev/#/pools/1655476167/assets

Happy to either create a new ticket or we use this one?

0x4Graham:
@kattylucy for this ticket, 

We should only remove the maturity date column if: 

All assets don't have a maturity date. 

If one has a maturity date, then we need to have the column.

0x4Graham:
Approved 👍

PR #2354, opened 2024-08-08
Bug asset name and maturity date

kattylucy:
- Date should be blank if null instead of "invalid date"
- Asset name was only displaying assetId

#2330 

### Approvals

- [ ] Dev


sophialittlejohn (approved):
LGTM

Issue #2312, opened 2024-07-25
Sync Token & NAV Prices: Pool Performance


0x4Graham:
There is a discrepancy in the displayed data for the NAV and token price. The values shown in red are sourced from the subquery, which lags. The green values come directly from the chain and are more accurate. To ensure the most accurate and up-to-date information is displayed, we should use the chain data for the fixed token price and the latest NAV.

![Image](https://github.com/user-attachments/assets/fa62cdde-a498-4739-9140-5dc775c8acef)

Solution: 

- Use the on-chain number for the fixed token price.

- For the latest item in the graph, utilize the most up-to-date chain number for Token Price & NAC

kattylucy:
I realize the currency is also different. do we want to update those?

in the graph is USD and on the button USDC. 

@0x4Graham @sonam-jo

0x4Graham:
Yes, they should both be USD

Issue #2310, opened 2024-07-25
Pool Overview: Performance Graph Buggy and doesn't have a set pattern


0x4Graham:
The pool performance chart has some weird behaviors when using. 

1. The x axis is not updating when selecting the date slider. In the following screenshot, it looks like the NAV for the 30 days starts at 0 and then increases 

![Image](https://github.com/user-attachments/assets/bd754c5d-98b7-48aa-bdef-fbd209d18d8c)

2. Months are skipped in some cases when playing around with the graphs/dates. 

![Image](https://github.com/user-attachments/assets/5892fb48-103b-4db4-992a-5d6f0780fc64)

Solution:  Can we bring some behaviour consistency into the charts on the pool performance chart.  

3. The NAV on chart is different to the NAV on the top and the blue box

![Image](https://github.com/user-attachments/assets/3b68dc39-dbcc-4afa-8cbb-1ff5c7328ba3)

kattylucy:
Here are the updates for this ticket when QAing 

Utilize on chain data for token price & nav
Keep only months on graph regarding data length to maintain consistency
Show only one month per internal, do not repeat month or skip months

PR #2356, opened 2024-08-08
Utilize on chain data for token price pool graph

kattylucy:
### Description

- Utilize on chain data for token price & nav 
- Keep only months on graph regarding data length to maintain consistency 
- Show only one month per internal, do not repeat month or skip months

#2312 
#2310 

### Approvals

- [ ] Dev

### DIFF
--- a/centrifuge-app/src/components/Charts/PoolPerformanceChart.tsx
+++ b/centrifuge-app/src/components/Charts/PoolPerformanceChart.tsx
@@ -115,15 +115,21 @@ function PoolPerformanceChart() {
   if (truncatedPoolStates && truncatedPoolStates?.length < 1 && poolAge > 0)
     return <Text variant="body2">No data available</Text>
 
-  const getXAxisInterval = () => {
-    if (rangeNumber <= 30) return 5
-    if (rangeNumber > 30 && rangeNumber <= 90) {
-      return 14
-    }
-    if (rangeNumber > 90 && rangeNumber <= 180) {
-      return 30
-    }
-    return 45
+  const getOneDayPerMonth = (): any[] => {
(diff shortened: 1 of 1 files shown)


sophialittlejohn (commented):
there still seems to be a discrepancy between the token prices displayed - I'm seeing three different token prices on the page (above the chart, in the chart tooltip for today, underneath the chart in the blue box).I think it would help to replace the last day of the subquery data in the chart with chain data from today?

kattylucy:
Hmmm that's weird. I will double check. It seems we want to move away from the subquery for this one.

Thank you for checking!! @sophialittlejohn

sophialittlejohn (commented):
I would suggest a slightly different approach:
```js
  const todayPrice = pool?.tranches[pool.tranches.length - 1].tokenPrice?.toDecimal().toNumber() || null
  const todayNav = pool.nav.total.toDecimal().toNumber()

  const isSingleTranche = pool?.tranches.length === 1
  const data: ChartData[] = React.useMemo(
    () =>
      truncatedPoolStates?.map((day) => {
        const nav = day.poolState.netAssetValue.toDecimal().toNumber()
        const price = (isSingleTranche && Object.values(day.tranches)[0].price?.toFloat()) || null
        // Replace the data as it's being created for the chart so that data stays consistent with the UI
        if (day.timestamp && new Date(day.timestamp).toDateString() === new Date().toDateString()) {
          return { day: new Date(day.timestamp), nav: todayNav, price: todayPrice }
        }
        return { day: new Date(day.timestamp), nav, price }
      }) || [],
    [isSingleTranche, truncatedPoolStates]
  )
```

sophialittlejohn (approved):
🥳

Issue #2331, opened 2024-07-30
LP v2: `CentrifugeRouter` integration


hieronx:
All investment interactions should now go through the `CentrifugeRouter` contract. 

- [ ] `lp.requestDeposit` => `CentrifugeRouter.multicall([open(), requestDeposit()])`
- [ ] `lp.requestDepositWithPermit` => `CentrifugeRouter.multicall([open(), permit(), requestDeposit()])`
- [ ] `lp.mint` => `CentrifugeRouter.claimDepositRequest`
- [ ] `lp.requestRedeem` => `CentrifugeRouter.requestDeposit`
- [ ] `lp.withdraw` => `CentrifugeRouter.claimRedeemRequest`
- [ ] `lp.cancelDepositRequest` => `CentrifugeRouter.cancelDepositRequest`
- [ ] `lp.claimCancelDepositRequest` => `CentrifugeRouter.claimCancelDepositRequest`
- [ ] `lp.cancelRedeemRequest` => `CentrifugeRouter.cancelRedeemRequest`
- [ ] `lp.claimCancelRedeemRequest` => `CentrifugeRouter.claimCancelRedeemRequest`

sophialittlejohn:
should these be the following instead of what's in the ticket above:
- [ ] lp.withdraw => CentrifugeRouter.claimRedeem
- [ ] lp.requestRedeem => CentrifugeRouter.requestRedeem
- [ ] lp.mint => CentrifugeRouter.claimDeposit

hieronx:
Yep @sophialittlejohn good catches!

PR #2360, opened 2024-08-09
LPv2 Centrifuge Router

sophialittlejohn:
### Description

This pull request...

- [x] lp.requestDeposit => CentrifugeRouter.multicall([enable(), requestDeposit()])
- [x]  lp.requestDepositWithPermit => CentrifugeRouter.multicall([enable(), permit(), requestDeposit()])
- [x]  lp.withdraw => CentrifugeRouter.claimRedeem
- [x]      lp.requestRedeem => CentrifugeRouter.requestRedeem
- [x]    lp.mint => CentrifugeRouter.claimDeposit
- [x]  lp.cancelDepositRequest => CentrifugeRouter.cancelDepositRequest
- [x]  lp.claimCancelDepositRequest => CentrifugeRouter.claimCancelDepositRequest
- [x]        lp.cancelRedeemRequest => CentrifugeRouter.cancelRedeemRequest
- [x]        lp.claimCancelRedeemRequest => CentrifugeRouter.claimCancelRedeemRequest

Adds gmp debug flag that will show a link to the message on axelar and will update when gas is paid and when tx is executed.

#2331
#2350 

### Approvals

- [ ] Dev


centrifuge-app/src/utils/useGmp.tsx
```diff
@@ -0,0 +1,143 @@
...
+        console.error('Error fetching data:', error)
+        return of([])
+      })
+    )
+
+  useQuery(
+    ['gmp'],
+    () => {
+      return new Promise((resolve) => {
+        const subscription = interval(5000)
```
onnovisser, line 63: this feels odd, having the interval here and the refetchInterval below. Wouldn't it be more straightforward without rxjs here?

centrifuge-js/src/modules/liquidityPools.ts
```diff
@@ -69,15 +70,35 @@ export function getLiquidityPoolsModule(inst: Centrifuge) {
     )
   }
 
+  function centrifugeRouter(chainId: number) {
+    const centrifugeRouter = getCentrifugeRouterAddress(chainId)
+    const getEstimate = from(contract(centrifugeRouter, ABI.CentrifugeRouter).estimate([0]))
+    return from(getEstimate).pipe(
+      switchMap((estimate) => {
+        return of({ estimate, centrifugeRouter })
+      })
```
onnovisser, line 79:
```suggestion
    return getEstimate.pipe(
      map((estimate) => {
        return { estimate, centrifugeRouter }
      })
```

centrifuge-app/src/utils/useGmp.tsx
sophialittlejohn: you're right, I cleaned it up and now it's simply returning a promise and useQuery is handling the refetching instead of an interval! Thanks!

centrifuge-app/src/utils/useGmp.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+export function useGmp() {
+  const { showGmp } = useDebugFlags()
+  const { isEvmOnSubstrate } = useWallet()
+  const centrifuge = useCentrifuge()
+  const isDemo = centrifuge.config.centrifugeWsUrl.includes('k-f.dev')
+  const axelarApiUrl = `https://${isDemo ? 'testnet' : ''}.api.axelarscan.io/gmp/searchGMP`
+  const axelarScanUrl = `https://${isDemo ? 'testnet' : ''}.axelarscan.io/gmp/`
+
+  const [gmp, setGmp] = useState<Gmp>(() => {
```
onnovisser, line 23: I would separate the txHash from the rest, as the txHash is a dependency of the query. And only the txHash needs to be in storage.

centrifuge-app/src/utils/useGmp.tsx
```diff
@@ -0,0 +1,131 @@
...
+      })
+      const result = await response.json()
+      return result.data
+    } catch (error) {
+      console.error('Error fetching data:', error)
+      return []
+    }
+  }
+
+  useQuery(
```
onnovisser, line 55: Then just take the data from the useQuery return, instead of setting the state in the query callback, which can lead to a bug if useGmp is used more than once on the page

centrifuge-app/src/utils/useGmp.tsx
sophialittlejohn: I've added the poolId and trancheId so that we know where the message belongs to render it appropriately, in fact I even though about adding a name to the storage object so that we know specifically for which transaction the message is for. Would you recommend tracking it some other way? The useQuery only returns information on the message, nothing about what triggered it

sophialittlejohn: But do you maybe mean to set the storage object after the query has completed rather than in the useQuery hook itself?

sophialittlejohn: I think I understood what you meant, now the storage object only holds the necessary information: txHash, poolId, trancheID and address. The status is captured directly from the useQuery data

onnovisser: Yes exactly!

onnovisser (approved):
💯

centrifuge-app/src/utils/useGmp.tsx
```diff
@@ -0,0 +1,126 @@
...
+        setGmp(JSON.parse(storedGmp))
+      }
+    }
+
+    if (executed) {
+      return () => {
+        sessionStorage.removeItem('gmp')
+      }
+    }
+  }, [gmp])
```
onnovisser, line 46: jsut one last lint warning

PR #2373, opened 2024-08-14
Support corrections (increase/decrease debt)

hieronx:
### Description

This pull request...

#<issue-number>

### Approvals

- [ ] Dev

### DIFF
--- a/centrifuge-app/src/components/PoolOverview/TransactionHistory.tsx
+++ b/centrifuge-app/src/components/PoolOverview/TransactionHistory.tsx
@@ -64,6 +64,11 @@ export const TransactionHistoryTable = ({
       return {
         label: 'Cash transfer',
         amount: transaction.amount,
+        netFlow: activeAssetId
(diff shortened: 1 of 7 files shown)


centrifuge-app/src/components/PoolOverview/TransactionHistory.tsx
```diff
@@ -64,34 +64,63 @@ export const TransactionHistoryTable = ({
       return {
```
kattylucy, line 64: Maybe is a good idea to have this line of code on line 63 

```
const netFlow = activeAssetId
          ? activeAssetId === transaction.toAsset?.id.split('-')[1]
            ? 'positive'
            : 'negative'
          : 'neutral',
```

and then add netFlow below in each corresponding case. In case we are debugging in the future so we can only change one place instead of the 5 different values within the same function.

hieronx: Agreed! Will change.

kattylucy (approved):
code wise looks good to me!

Issue #2364, opened 2024-08-12
Display EVM addresses


hieronx:
Wherever we display addresses, we should detect if they are EVM addresses, and if so, display the EVM format.

Note: it is correctly implemented in the `Oracle providers` section already, logic can be copied from there.

Example how it should look:
<img width="333" alt="Screenshot 2024-08-12 at 13 39 12" src="https://github.com/user-attachments/assets/f9312bd0-2585-44c6-8a0e-e2bdcbf543c6">

### TODOs
- [x] Pool Manager addresses
<img width="335" alt="Screenshot 2024-08-12 at 13 37 14" src="https://github.com/user-attachments/assets/1d433bac-47ca-410c-ae9f-0f479e6409c1">

- [ ] Asset originator addresses
<img width="488" alt="Screenshot 2024-08-12 at 13 37 32" src="https://github.com/user-attachments/assets/caca774e-eb6f-43d1-be41-75d81f6e4a47">

PR #2374, opened 2024-08-14
Display evm address

kattylucy:
Display EVM address if EVM instead of converting and displaying the substrate address

#2364 

### Approvals

- [ ] Dev


hieronx:
Just tested this a bit locally, works perfectly!

hieronx (approved):

Issue #2383, opened 2024-08-19
Reporting Tab in "Anemoy DeFi Yield Fund 1 SP" a little buggy with loading


0x4Graham:
When navigating to the "reporting" tab within "Anemoy DeFi Yield Fund 1 SP" sometimes it works and other times an error asking to try again is displayed 

![Image](https://github.com/user-attachments/assets/57560d45-cb5a-4034-84ff-7af3f9a093ce)

![Image](https://github.com/user-attachments/assets/9837e1b5-3946-4bad-9f07-41f4ed5c4b90)

PR #2393, opened 2024-08-20
Fix date for reporting

kattylucy:
Date should always be date type before submitting to subquery to avoid errors 

#2383 

- [ ] Dev


sophialittlejohn:
Not sure if this was happening before this PR already but I just left the default date value and changed monthly to daily and am seeing the date displayed off by one. Regardless, I think it would be great to fix as part of this PR 😄 

Screenshot 2024-08-21 at 16 50 39

sophialittlejohn (approved):
👍

Issue #2164, opened 2024-05-29
Multiple RPC endpoints


sophialittlejohn:
We should consider having our app be multi-endpoint, ideally selecting the first available working endpoint from a list and also having a drop-down allowing the user to change manually

### Tasks:
- [ ] automatically switch RPC endpoint if the current one is not reachable
- [ ] display error message if all of the RPC endpoints can't be reached

### TBD

1. find all possible RPC endpoints to use
2. define criteria to determine whether the RPC endpoint is reachable or not

sophialittlejohn:
Blocked until cent-js rewrite

sophialittlejohn:
Unblocking due to the many issues of late with our RPC providers. For evm chains we should use etherjs [getDefaultProvider](https://docs.ethers.org/v5/api/providers/#providers-getDefaultProvider)

PR #2394, opened 2024-08-20
Multiple RPC endpoint support

sophialittlejohn:
### Description

This pull request...

- use the provider from `getDefaultNetwork` for centrifuge-react connectors
- clean up leftover from goerli
- fix warnings in console
- fix start script in main package.json
- accepts a comma separated list of rpc urls for polkadot and connects to the first healthy one
- upgrades ethersjs to v6

[2164](https://github.com/centrifuge/apps/issues/2164)

### Approvals

- [ ] Dev


centrifuge-react/src/components/WalletProvider/WalletProvider.tsx
```diff
@@ -240,11 +240,26 @@ export function WalletProvider({
 
   const [proxies] = useCentrifugeQuery(['allProxies'], (cent) => cent.proxies.getAllProxies())
 
-  function getProvider(chainId: number) {
-    return (
-      cachedProviders[chainId] ||
-      (cachedProviders[chainId] = new JsonRpcProvider((evmChains as any)[chainId].urls[0], chainId))
-    )
+  async function findHealthyProvider(chainId: number, urls: string[]): Promise<JsonRpcProvider> {
+    for (const url of urls) {
```
onnovisser, line 244: Would use a Promise.any here

centrifuge-react/src/components/WalletProvider/WalletProvider.tsx
```diff
@@ -240,11 +240,26 @@ export function WalletProvider({
...
+    }
+    throw new Error(`No healthy provider found for chain ${chainId}`)
+  }
+
+  function getProvider(chainId: number): JsonRpcProvider {
+    const urls = (evmChains as any)[chainId].urls
+    if (!cachedProviders.has(chainId)) {
+      findHealthyProvider(chainId, urls).catch(console.error)
+    }
+    return cachedProviders.get(chainId) || new JsonRpcProvider(urls[0], chainId)
```
onnovisser, line 262: with `findHealthyProvider` being async, this may return a `new JsonRpcProvider` multiple times and not garantee that it's healthy

sophialittlejohn: I've changed the implementation to now use ethers [`getDefaultProvider()`](https://docs.ethers.org/v6/api/providers/#getDefaultProvider) its, much cleaner and doesn't require the manual work from before

centrifuge-app/src/utils/tinlake/useTinlakeBalances.ts
```diff
@@ -31,10 +30,10 @@ export function useTinlakeBalances(address?: string) {
 
 const WCFG_ADDRESS = '0xc221b7e65ffc80de234bbb6667abdd46593d34f0'
 
-async function getBalances(pools: TinlakePool[], address: string, provider: JsonRpcProvider) {
+async function getBalances(pools: TinlakePool[], address: string, provider: Provider) {
   const calls: EvmMulticallCall[] = []
-  const toTokenBalance = (val: BigNumber) => new TokenBalance(val.toString(), 18)
-  const toCurrencyBalance = (val: BigNumber) => new CurrencyBalance(val.toString(), 18)
+  const toTokenBalance = (val: BigInt) => new TokenBalance(val.toString(), 18)
```
onnovisser, line 35: BigInt as a type has to be lowercase everywhere

centrifuge-js/src/CentrifugeBase.ts
```diff
@@ -295,19 +296,69 @@ type Events = ISubmittableResult['events']
 
 const txCompletedEvents: Record<string, Subject<Events>> = {}
 const blockEvents: Record<string, Observable<Events>> = {}
+let parachainUrlCache: string | null = null
```
onnovisser, line 299: this won't work if there's different instances of the Centrifuge class. It's why the above cached are keyed by parachain url. I think it would be better to keep this as a property on the class. It would need to be copied over when the instance gets cloned though, like in `Centrifuge->connect()`

centrifuge-js/src/CentrifugeBase.ts
```diff
@@ -459,7 +510,13 @@ export class CentrifugeBase {
         }),
         tap((result) => {
```
onnovisser, line 511: The `tap` callback can be an async function and then `this.getTxCompletedEvents()` can simply be awaited

centrifuge-js/src/CentrifugeBase.ts
```diff
@@ -459,7 +510,13 @@ export class CentrifugeBase {
...
           options?.onStatusChange?.(result)
-          if (result.status === 'InBlock') this.getTxCompletedEvents().next(result.events)
+          if (result.status === 'InBlock') {
+            from(this.getTxCompletedEvents())
+              .pipe(take(1))
+              .subscribe((subject) => {
+                subject.next(result.events)
+              })
```
onnovisser, line 518:
```suggestion
            (await this.getTxCompletedEvents()).next(result.events)
```

centrifuge-js/src/modules/tinlake.ts
```diff
@@ -94,6 +93,7 @@ export function getTinlakeModule(inst: Centrifuge) {
   ) {
     const [tranche] = args
     return pending(
+      // @ts-ignore
```
onnovisser, line 96: what are all these errors saying?? 😅😅😅

centrifuge-js/src/CentrifugeBase.ts
```diff
@@ -295,19 +296,69 @@ type Events = ISubmittableResult['events']
...
+        clearTimeout(timer)
+        ws.close()
+        resolve(true)
+      }
+
+      ws.onerror = () => {
+        clearTimeout(timer)
+        ws.close()
+        console.log(`Connection to ${url} failed`)
+        resolve(false)
```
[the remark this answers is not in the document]
sophialittlejohn: the only issue with rejecting the promise is that it logs an error to the console which I was trying to avoid, I'd rather control the error handling myself and resolving to false allows me to do so

centrifuge-js/src/CentrifugeBase.ts
onnovisser: what about making this a `Set` `reachableParachainUrls`? Then when getting the cached url, you it can check which of the config urls is in the set. Then it still works if there's multiple instances of Centrifuge with different configs

centrifuge-js/src/modules/tinlake.ts
sophialittlejohn: The types just weren't inferred properly by the connected contract. When using the the ethers `Contract` directly the methods on the contract are not type checked, but when using a connected contract the are type check even though the method types seem to be non-inferable from the ABI.

centrifuge-js/src/CentrifugeBase.ts
```diff
@@ -295,19 +296,66 @@ type Events = ISubmittableResult['events']
...
+        ws.close()
+        console.log(`Connection to ${url} failed`)
+        reject()
+      }
+    })
+  }
+
+  private async getCachedParachainUrl(): Promise<string> {
+    if (parachainUrlCache.values().next().value) {
+      return parachainUrlCache.values().next().value
```
onnovisser, line 351:
```suggestion
    const match = this.rpcEndpoints.find(url => parachainUrlCache.has(url))
    if (match) {
      return match
```

onnovisser:
@sophialittlejohn Looking good! One last comment.

onnovisser (approved):

Issue #2315, opened 2024-07-25
Invest/Redeem Slider: Allow TxHistory to open in new page


0x4Graham:
When you're on the Invest/Redeem side bar, and click on Transaction History it navigates to Transaction History page. We don't want this to happen as it can result in users losing their state, if they were busy with an investment. 

Tx History should open in a new page.

PR #2398, opened 2024-08-21
Open view all transactions on new tab

kattylucy:
- New tab should open instead of redirecting on page (view all transactions, invest drawer component)

#2315

- [ ] Dev


sophialittlejohn (approved):
💯

Issue #2377, opened 2024-08-15
Deploying LP tranches doesn't work if connected with EVM wallet as pool admin


hieronx:
If you are connected with an EVM wallet as the pool admin, and you click `Switch to Ethereum Sepolia`, the page goes blank. It only works if the pool admin you are connected with is a non EVM wallet.

sophialittlejohn:
Show page regardless of admin status.

PR #2400, opened 2024-08-22
Fix deploying LP tranches with EVM wallet

onnovisser:
### Description

If you are connected with an EVM wallet as the pool admin, and you click Switch to Ethereum Sepolia, the page goes blank, because the account no longer registers as a pool admin. Remember whether the user was admin at some point while on the Investors tab to keep the tab visible

#2377

### Approvals

- [ ] Dev


sophialittlejohn (approved):
😅

Issue #2390, opened 2024-08-20
Pool Overview Page Redesign


0x4Graham:
The following ticket will cover all the redesign task items for the Pool Overview Page Redesign/Improvements 

**Please ensure you update the component library(story book) for all existing & new components**

https://www.figma.com/design/ng7qdNcSCXSDA6ZUdWIs6u/Pool-Overview%2F-Pool-Detail?node-id=2173-41892&t=UbtuoHdsKv1ZIAVQ-0

- [ ] Title of page changes to Pools of real-world assets in semi bold size 16

- [ ] Change the colour of the sidebar #252B34. When tab selected or hovered over, change the colour of the text in the menu bar to: #FFC500. Ensure that we update the story book components for this as well. 

![Image](https://github.com/user-attachments/assets/2cd4658f-f7e8-4f82-a93b-37e33b3b9d3d)

- [ ] Button colour changes. 

- For connect wallet:  #252B34
- For secondary buttons: yellow
- For primary: #252B34

- [ ] Remove the graph at the top of the page. Colour of the values text needs to be:  #37394C and the headings: #82888D   (Need to ensure that we're able to query the values)
- Total Value Locked (TVL) with small YoY growth %

![Image](https://github.com/user-attachments/assets/ec59cadc-130a-4f09-907d-285f9b4e33e5)

**For Archive Pools**

- [ ] Change font colour for "View Archived pools" to 82888D
- [ ] Remove underline and replace with > 

![Image](https://github.com/user-attachments/assets/896f7669-dbc7-48b1-978b-df9f81bbaee0)

kattylucy:
Can I get a design for the button "create a pool" 

@sonam-jo @BrianCarter-Design 

![Image](https://github.com/user-attachments/assets/4bc01830-052e-4929-9816-4fcd29a677d9)

sonam-jo:
This doesn't fall under the primary/secondary colour system? Then let's do the background grey 82888D and the text colour 252B34

kattylucy:
You are right, it does fall under primary/secondary. @sonam-jo

0x4Graham:
@kattylucy This PR breaks the transaction history page at the top 

![Image](https://github.com/user-attachments/assets/1da155cc-aeb1-4bd3-9a44-3b1427b1f0fc)

The button is overlaying on the table

kattylucy:
https://app-dev.k-f.dev/#/history @0x4Graham

BrianCarter-Design:
[...]

PR #2402, opened 2024-08-22
Pool overview page redesign

kattylucy:
### Description

- Redesign menu on the left and primary/secondary colors for the buttons
- Redesign layout and router to use outlet 
- Check on storybook components
- Delete alternative color theme (dark / light ) 

#2390

### Approvals

- [ ] Dev


sophialittlejohn (commented):
Nice work @kattylucy 🔥 

The Banner that's sometimes displayed in the bottom right corner could most definitely benefit from a style upgrade. What do you think about a black background with white text? 

In general do you think it would make sense to replace the primary color in the theme for gold to ease the transition?

fabric/src/theme/tokens/theme.ts
```diff
@@ -29,6 +34,7 @@ const lightColors = {
   backgroundInput: 'white',
   backgroundThumbnail: grayScale[500],
   backgroundInverted: grayScale[900],
+  backgroundBlack: black,
```
onnovisser, line 37: I would change the `backgroundInverted` to be `black` and use that for the sidebar

onnovisser: or better yet, change grayScale[900] to be `#252B34`.

fabric/src/theme/tokens/theme.ts
```diff
@@ -14,11 +14,16 @@ const statusWarningBg = yellowScale[50]
...
+const black = '#252B34'
+const gold = '#FFC500'
+
+const colors = {
   textPrimary: grayScale[900],
   textSecondary: grayScale[800],
   textDisabled: grayScale[500],
   textInverted: 'white',
+  textGold: gold,
+  textBlack: black,
```
onnovisser, line 26: I would just use textPrimary. If the primary text color has change, we should update that instead

centrifuge-app/src/utils/useListedPools.ts
```diff
@@ -40,3 +48,83 @@ export function useListedPools() {
...
+  const flattenedData =
+    data?.pools?.nodes.flatMap((pool: any) =>
+      pool.poolSnapshots.nodes.map((snapshot: any) => ({
+        netAssetValue: snapshot.netAssetValue,
+        decimals: pool.currency.decimals,
+      }))
+    ) || []
+
+  // Aggregate NAV from last year
+  const aggregatedNetAssetValue = flattenedData.reduce((accumulator: any, item: FlattenedDataItem) => {
```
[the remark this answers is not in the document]
sophialittlejohn: you'll have to import it 
```js
import Decimal from 'decimal.js-light'
```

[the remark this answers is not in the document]
kattylucy: Oh i missed this change. THANK YOU!

PR #2448, opened 2024-09-12
misc: Adapt docs and use correct reserve amount

mustermeiszer:
Just fixing using `pool.reserve.total` instead of `pool.reserve.available`.

Minors: 
* added docs for the future
* removed redundant calculation between pannel and card

### DIFF
--- a/centrifuge-app/src/pages/NavManagement/NavManagementAssetTable.tsx
+++ b/centrifuge-app/src/pages/NavManagement/NavManagementAssetTable.tsx
@@ -195,6 +195,11 @@ export function NavManagementAssetTable({ poolId }: { poolId: string }) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [initialValues, isEditing, isLoading])
 
+
+  // NOTE: This assumes that pool.reserve.total comes from onchain state AND NOT from the runtime-apis
+  const totalAum = pool.nav.aum.toDecimal().add(pool.reserve.total.toDecimal())
+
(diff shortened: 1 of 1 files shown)


centrifuge-app/src/pages/NavManagement/NavManagementAssetTable.tsx
```diff
@@ -195,6 +195,11 @@ export function NavManagementAssetTable({ poolId }: { poolId: string }) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [initialValues, isEditing, isLoading])
 
+
+  // NOTE: This assumes that pool.reserve.total comes from onchain state AND NOT from the runtime-apis
```
mustermeiszer, line 199: @onnovisser or @sophialittlejohn is this correct? Does the `pool.reserve.total` come from onchain state? That is critical here.

sophialittlejohn: the value you're using definitely comes from onchain 
https://github.com/centrifuge/apps/blob/15ac6108c79468ce668f1f1ff983f53bef10595b/centrifuge-js/src/modules/pools.ts#L2092

mustermeiszer: Perfect. As long as it is not coming from `api.PoolsApi.nav()` we are good.

mustermeiszer:
Can we merge this then?

onnovisser (approved):
LGTM

PR #2449, opened 2024-09-12
Fix average maturity

onnovisser:
### Description

Exclude cash assets and assets without maturity from the calculation, and round at the end.

### Approvals

- [ ] Dev


kattylucy (approved):
👍

Issue #2432, opened 2024-09-06
Investigate huge number of calls to Subquery


onnovisser:
Noticed over 800 calls to Subquery happening after closing an epoch. Investigate why there are so many and what we can do to reduce them

PR #2454, opened 2024-09-17
Reduce calls to subquery

onnovisser:
### Description

Getting the latest tranche snapshots for all tranches for all pools was the cause of the large amount of calls to subquery. There was a call for each tranche, and it was done more than once for some reason. With over a hundred pools/tranches on demo that was a lot. Now getting the yields in `getPools` is done with one call for all pools/tranches.

[#2432](https://github.com/centrifuge/apps/issues/2432)

### Approvals

- [ ] Dev


sophialittlejohn (approved):
Nice work!

Issue #2548, opened 2024-11-28
Prime Page: Redesign


0x4Graham:
The redesign is a continuation of #2547

https://www.figma.com/design/ng7qdNcSCXSDA6ZUdWIs6u/Pool-Overview%2F-Pool-Detail?node-id=4229-28053&node-type=canvas&t=CXmTnlqvxTUarwyc-0

The main difference's being:
1. The graph will take up the entire width and we won't show the block with assets to invest. 
2. There will be no CTA on the _Investment Positions_ table

sonam-jo:
More details

**On the Prime page**
- only cosmetic changes, see the graphic added on the right side. Portfolios title for the table. 

**Gnosis DAO page**
- Summary box on the top like the Assets pages with `Portfolio value`, `% increase` in value since the investment was made for the first time, `realised P&L` and `unrealised P&L`. Note, this would be a total sum value, if there were multiple investments. 
- Change format of the chart, similar to the one we have on the Overview page. It should show the total sum of the portfolio value (including if there were multiple investments). 
- Holdings will be renamed to `Investment positions` with the same information as today, but also include `Network`, Realised P&L and unrealized P&L. 
- No changes to the Transaction table, but note the current CTA for View all is incorrectly linked. When the user clicks on View all, it should now populate all the transactions within the same table, with the Resolutions section shifting down. 
- For Resolutions, note the change in graphics - rest of the content is same. We have added a new logic for View more CTA at the bottom. Whenever there are more than 4 resolutions, we should have a View more CTA. When the user clicks on it, it should open the next 4 resolutions in a second row. The View more CTA is not visible when all the resolutions have been populated.

sonam-jo:
Looks good Katty. Small cosmetic changes

- Add `USD` to the `Portfolio value` in the box and in the chart

![Image](https://github.com/user-attachments/assets/e4dbaa11-5326-4f3f-bc11-4b5d29fc1d99)

- Change colour for `Download` to light grey like the `View all` button

![Image](https://github.com/user-attachments/assets/0df2163d-f4b7-4c62-816e-99a57a7a9710)

PR #2554, opened 2024-12-13
Prime page - redesign

kattylucy:
Redesign prime page 

https://www.figma.com/design/ng7qdNcSCXSDA6ZUdWIs6u/Pool-Overview%2F-Pool-Detail?node-id=4229-28053&p=f&t=8l0I5rUdl6HZZynm-0

#2548 

- [ ] Dev


centrifuge-app/src/components/BackButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <StyledRouterLinkButton to={to} small icon={<IconArrowLeft />} variant="tertiary" />
+      <Box display="flex" alignItems="center">
+        <Text variant="heading1" style={{ marginRight: 8 }}>
+          {label}
+        </Text>
+        {children}
+      </Box>
+    </Box>
+  )
+}
```
sophialittlejohn, line 54: I think this would fit well into fabric, maybe pass the `RouterLinkButton` as a prop?

fabric/src/theme/types.ts
```diff
@@ -27,7 +27,8 @@ type BackgroundColorName = `background${
   | 'Thumbnail'
   | 'AccentPrimary'
   | 'AccentSecondary'
-  | 'Inverted'}`
+  | 'Inverted'
+  | 'Light'}`
```
sophialittlejohn, line 31: It looks like you meant to add a new color to the border colors instead of background? Also any chance you could update either primary or secondary of the existing border colors instead of adding a new one? `Light` doesn't really fit into our naming semantics as the colors are typically defined with how they're used rather than what they are

sophialittlejohn (changes requested):
Great job! I left a few comments for improvements for the code. And also have a few improvement suggestions for the chart
1. y-axis labels: the values and their precision should depend on the how big the gaps are in between. If you look at the 30 days filter the values appear to be too close together.
2. x-axis labels: also here we should define the label differently depending on how large the filtered dataset is. For 30 days I think we should label every day, for 90 days possibly every 3rd, anything bigger just display the month. If the data set is only 30 days total but the filter is set to 90 days, it should still show every day on the x-axis label. 
3. The chart fills in after the page has loaded, we should find a more elegant solution than letting the content jump. Reserving the space in advance is an option, also figuring out how to load the chart in advance could be an option.
Screenshot 2025-01-17 at 17 03 33

centrifuge-app/src/components/Portfolio/Holdings.tsx
```diff
@@ -139,6 +180,7 @@ export function useHoldings(address?: string, chainId?: number, showActions = tr
       ...token,
       tokenPrice: token.tokenPrice.toDecimal() || Dec(0),
       showActions,
+      connectedNetwork: token.chainId,
```
sophialittlejohn, line 183: Looks better! Now I think you just need to update the name `connectedNetwork` to `chainId`

centrifuge-app/src/components/Portfolio/Holdings.tsx
```diff
@@ -251,8 +294,19 @@ export function Holdings({
...
         onClose={() => navigate(pathname, { replace: true })}
       />
-      <DataTable columns={columns} data={tokens} defaultSortKey="position" />
-    </>
+      <DataTable
+        headerStyles={{
+          backgroundColor: 'white',
+          border: 'transparent',
+          borderBottom: `1px solid ${theme.colors.backgroundInverted}`,
+        }}
```
sophialittlejohn, line 302: I'm starting to see these table styles more often. I think we should refactor this usage (and all other overwriting of the headerstyles) into a table variant

centrifuge-app/src/components/BackButton.tsx
sophialittlejohn: I also saw this type of back arrow used on the asset detail page, please make sure they're using the same base component

centrifuge-app/src/components/Portfolio/Holdings.tsx
```diff
@@ -35,7 +35,11 @@ export type Holding = {
   tokenPrice: Decimal
   showActions?: boolean
   address?: string
-  connectedNetwork?: string | null
+  chainId?: number
+  realizedProfit?: CurrencyBalance
+  unrealizedProfit?: CurrencyBalance
+  yieldSinceInception?: Perquintill | null
+  connectedNetwork?: any
```
sophialittlejohn, line 42: oops, probs meant to remove this?

sophialittlejohn (approved):
🚢 🚢 🚢

Issue #2561, opened 2024-12-18
Connect wallet feedback


hieronx:
- [x] Only the arrow icons are clickable, it would be better if the full box was clickable:
<img width="544" alt="Screenshot 2024-12-18 at 13 28 10" src="https://github.com/user-attachments/assets/57578169-d994-4314-8ab3-1aaacab8e230" />

- [x] If you are already connected, and you click Switch wallet, it opens like this:
<img width="640" alt="Screenshot 2024-12-18 at 13 29 15" src="https://github.com/user-attachments/assets/8e1726f1-c430-4133-aa7b-3679a4f3ef3d" />

In this case I would expect the wallet box to already be expanded.

PR #2562, opened 2024-12-18
Updates connect wallet button

kattylucy:
- Make whole box clickable and not only arrow icon
- If you are already connected, the dropdown should be open

#2561


hieronx:
Thanks for the quick improvements! Noticed one more thing: if you have not yet selected a Network, the Wallet box should be disabled, since it is still empty.

Screenshot 2024-12-18 at 16 10 23

hieronx (approved):
Might want to change the styling in a future PR, see Slack message, but let's ship this as is 🚢

Issue #2109, opened 2024-05-02
Create a new data report “Orders"


annamehr:
Create a new list in Data tab called `Orders`. This should cover the main metrics when **closing epochs**. Time period: per epoch. No data aggregation.

The following metrics should be part of it
* Epoch
* Date and time
* NAV 
* NAV per share (formerly token prices)
* Investments locked
* Investments executed
* Redemptions locked
* Redemptions executed  
* Paid fees
* Pending fees

sophialittlejohn:
waiting for #2094 to be merged before starting on this

kattylucy:
Notes for development: 

* The idea would be that you add an additional report to the reports tab on a pool (https://app.centrifuge.io/pools/4139607887/reporting)
* The data should not be aggregated meaning you should display the value on the current day
* The name of the report should be Orders
* The fees have their own hook usePoolFeeStates. the rest should probably come from the usePoolStates  hook

PR #2578, opened 2025-01-14
Add orders report

kattylucy:
- Add new order reports 

#2109

### DIFF
--- a/centrifuge-js/src/modules/pools.ts
+++ b/centrifuge-js/src/modules/pools.ts
@@ -3987,7 +3987,8 @@ export function getPoolsModule(inst: Centrifuge) {
       combineLatestWith(getPoolCurrency([poolId])),
       map(([data, poolCurrency]) => {
         return data?.epoches?.nodes.map((order) => {
-          const index = order.epochStates.nodes.length - 1
+          const index = order.epochStates.nodes.length > 1 ? order.epochStates.nodes.length - 1 : 0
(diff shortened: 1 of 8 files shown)


mustermeiszer:
Could we make the ordering descending from `Date & Time` the default?
image

mustermeiszer:
Some numbers are missing. Is this an indexing issue?

mustermeiszer:
Would remove the date here.

mustermeiszer (approved):
Approving from the product side. Should still wait code-wise for Sophia or Onno. But thanks, that is really, really helping a lot! 

Regarding the inconsistent data I will discuss with Jeroen and Graham. Not relevant for this PR.

sophialittlejohn (approved):
Looks good! Nice work Katty 🔥

centrifuge-js/src/modules/pools.ts
```diff
@@ -3941,6 +3942,85 @@ export function getPoolsModule(inst: Centrifuge) {
...
+                sumOutstandingInvestOrders
+                sumFulfilledInvestOrders
+                sumOutstandingRedeemOrders
+                sumFulfilledRedeemOrders
+              }
+            }
+            poolSnapshots{
+              nodes{
+                netAssetValue
+              }
```
mustermeiszer, line 3974: @hieronx this value is off from what we need. 

It seems to be the value without subtracting the fees, but I am not sure. Can you guide me to where it is coming from and whether we can use a more up-to-date value or better, the one from the API?

PR #2579, opened 2025-01-16
add new sdk to current app

kattylucy:
Integrate new sdk into out current app (reports and data tabs only)

- [ ] Dev

### DIFF
--- a/centrifuge-app/package.json
+++ b/centrifuge-app/package.json
@@ -29,7 +29,7 @@
     "@centrifuge/centrifuge-js": "workspace:*",
     "@centrifuge/centrifuge-react": "workspace:*",
     "@centrifuge/fabric": "workspace:*",
-    "@centrifuge/sdk": "^0.0.0-alpha.0",
+    "@centrifuge/sdk": "^0.0.0-alpha.8",
     "@makerdao/multicall": "^0.12.0",
     "@polkadot/react-identicon": "~3.1.4",
     "@styled-system/css": "^5.1.5",
--- a/centrifuge-app/package.json
+++ b/centrifuge-app/package.json
@@ -29,7 +29,7 @@
     "@centrifuge/centrifuge-js": "workspace:*",
     "@centrifuge/centrifuge-react": "workspace:*",
     "@centrifuge/fabric": "workspace:*",
-    "@centrifuge/sdk": "^0.0.0-alpha.11",
+    "@centrifuge/sdk": "latest",
     "@makerdao/multicall": "^0.12.0",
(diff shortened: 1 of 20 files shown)


centrifuge-app/src/utils/formatting.ts
```diff
@@ -80,3 +81,66 @@ export function truncateText(txt: string, len: number) {
   }
   return txt
 }
+
+// FORMAT FUNCTIONS FOR NEW SDK
+
+export function formatDecimal(decimalVal: any, displayDecimals: number, currency?: string): string {
```
sophialittlejohn, line 87:
```suggestion
export function formatDecimal(  amount: CurrencyBalance | TokenBalance | Price | Rate | Decimal | number,
, displayDecimals = 2, currency?: string): string {
```
Would love to see this function types like the other formatting options we had. There will be some type conflicts with Rate/Perquintill but I think you can circumvent the problem by writing the functions in a new file e.g `formatting-sdk.ts`. Then we can likely also keep the names `formatBalance`, `formatBalanceAbbreviated` and `formatPercentage` (I don't think `formatDecimal` really conveys what the function does)

centrifuge-app/package.json
```diff
@@ -29,7 +29,7 @@
     "@centrifuge/centrifuge-js": "workspace:*",
     "@centrifuge/centrifuge-react": "workspace:*",
     "@centrifuge/fabric": "workspace:*",
-    "@centrifuge/sdk": "^0.0.0-alpha.0",
+    "@centrifuge/sdk": "latest",
```
sophialittlejohn, line 32: Let's pin down the version so we don't accidentally include updates that we're not ready for yet

centrifuge-app/src/components/Report/FeeTransactions.tsx
```diff
@@ -58,14 +69,16 @@ export function FeeTransactions({ pool }: { pool: Pool }) {
     }
 
     return transactions
-      ?.filter((tx) => tx.type !== 'PROPOSED' && tx.type !== 'ADDED' && tx.type !== 'REMOVED')
+      ?.filter(
+        (tx) => tx.transactionType !== 'PROPOSED' && tx.transactionType !== 'ADDED' && tx.transactionType !== 'REMOVED'
+      )
```
sophialittlejohn, line 74: could we move this filter to the sdk?

sophialittlejohn (approved):
Amazing!!! 🔥🔥🔥 

Let's get this merged after this current release to give product enough time to test it before it goes live

kattylucy:
No problem. I will merge it into your reports branch and we can release from there!

Issue #2577, opened 2025-01-13
Enable Centrifuge Router on redemptions (if not already enabled)


sophialittlejohn:
We have programmed all investing and redeeming to go through the CentrifugeRouter contract. Prior to investing `enable()` is called. If the user invests with the centrifuge router they will also be able to redeem with it. However there are some edge cases where a user may have made an investment before the centrifuge router was installed and now want to make a redemption. For those users we have to called `enable` first.

- before a redemption check on the CentrifugeRouter is `isEnabled` is false
- if false call `enable()` as a first transaction and then claimRedeem in the second

PR #2580, opened 2025-01-16
Enable CentrifugeRouter if needed

onnovisser:
### Description

Enable the CentrifugeRouter before doing any other LP action, if the router isn't yet enabled for the user, which can happen for investments made before the router was implemented.

[#2577](https://github.com/centrifuge/apps/issues/2577)

### Approvals

- [ ] Dev


sophialittlejohn (approved):
Logic and code looks sound to me!

Issue #2526, opened 2024-11-01
Asset transactions table update


sonam-jo:
Data --> Asset transactions table 

![Image](https://github.com/user-attachments/assets/a03b1614-f8fe-4cd1-892c-d8076faf03d2)

Can we change the format of this table to be the same as the `Transaction history` table on the `Overview` page with:
- Transaction date
- Transaction
- Amount
- Epoch (we keep this from the current asset transactions table)
- View transaction

![Image](https://github.com/user-attachments/assets/31b49c0e-4651-49d0-a3e6-410197c8c1d3)

0x4Graham:
@kattylucy Is it possible to know what asset was purchased? Like we currently have in the transaction history table? I feel we should extend this to include the asset.  @sonam-jo what do you think? 

<img width="932" alt="Image" src="https://github.com/user-attachments/assets/4d2b2fd8-c639-4162-aed0-2f71df130fbd" />

sonam-jo:
Yes, similar to how we have the table in Transaction history on the Overview page, we need to have `Purchase of X asset`, `Sale of x asset`, `Financing of Y asset` under the `Transaction` column

kattylucy:
<img width="1061" alt="Image" src="https://github.com/user-attachments/assets/a3d5698d-035a-47da-8eea-49e483b9787b" />

kattylucy:
The filters not longer match the table anymore - which filters should we add or do I just remove it  and keep from/to @sonam-jo

PR #2584, opened 2025-01-20
Modify asset transactions table

kattylucy:
- Modify the asset transactions table inside data tab to be similar to how we have the table in Transaction history on the Overview page. We need to have Purchase of X asset, Sale of x asset, Financing of Y asset under the Transaction column

#2526

### DIFF
--- a/centrifuge-app/src/components/Report/AssetTransactions.tsx
+++ b/centrifuge-app/src/components/Report/AssetTransactions.tsx
@@ -1,59 +1,99 @@
-import { Pool } from '@centrifuge/centrifuge-js'
+import { CurrencyBalance, Pool } from '@centrifuge/centrifuge-js'
 import { useGetExplorerUrl } from '@centrifuge/centrifuge-react'
 import { Box, IconAnchor, IconExternalLink, Text } from '@centrifuge/fabric'
 import * as React from 'react'
 import { formatBalance } from '../../../src/utils/formatting'
+import { useBasePath } from '../../../src/utils/useBasePath'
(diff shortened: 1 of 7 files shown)


centrifuge-app/src/components/Report/AssetTransactions.tsx
```diff
@@ -64,75 +105,60 @@ export function AssetTransactions({ pool }: { pool: Pool }) {
...
+          'Transaction Date': `"${formatDate(transaction.timestamp, {
+            year: 'numeric',
+            month: 'long',
+            day: 'numeric',
+            hour: 'numeric',
+            minute: 'numeric',
+            second: 'numeric',
+            timeZoneName: 'short',
+          })}"`,
+          Transaction: `${import.meta.env.REACT_APP_SUBSCAN_URL}/extrinsic/${transaction.hash}`,
```
sophialittlejohn, line 147: using the subscan URL will only work if the network on which the transaction was made is `centrifuge`. We'll need to point to etherscan/basescan if the transaction was made on a different network. You should be able to fetch the network id from subquery

sophialittlejohn (commented):
Nice job! The filters need to be adapted for this report type. When filtering assets, it's probably simple enough to filter any report row where the selected asset type is mentioned in the `Transaction` column. The transaction type filter doesn't really make sense with the current wording anymore, maybe just remove it? Best to consult product if they have any other/different filters they want.

sophialittlejohn (approved):
LGTM 🔥

Issue #2566, opened 2024-12-20
Attestation EVM signing bug


hieronx:
When submitting a NAV update using an EVM wallet, the second step (where the batch transaction is submitted) fails:

![Screenshot 2024-12-20 at 10 24 09 AM](https://github.com/user-attachments/assets/1d3519db-1dac-4268-ac97-a5b71421f655)

It does work with a substrate wallet.

PR #2605, opened 2025-01-27
Fix EVM gas limit

onnovisser:
### Description

When the calculated gas amount exceeds the max gas limit, it doesn't get set at all and the transaction might fail as a result. 

[#2566](https://github.com/centrifuge/apps/issues/2566)

### Approvals

- [ ] Dev


sophialittlejohn (approved):
Hopefully this solves it! 🙏

Issue #2590, opened 2025-01-21
Auto-swap on repayment


hieronx:
In the repayment flow:
- If the pool currency is local USDC
- Retrieve list of currencies that can swap to this (lpEthUSDC etc)
- Sum up all the balances as the total USDC balance, and show this
- Automatically call tokenMux.deposit/burn in the batch for repaying

PR #2608, opened 2025-01-29
Repay form: Auto-swap currencies

onnovisser:
### Description

Adds a feature to the repay form to check for variants that can be exchanged for the pool currency via the token mux pallet and automatically swap them.

#2590

### Approvals

- [ ] Dev


centrifuge-app/src/pages/Loan/MuxRepay.tsx
sophialittlejohn, line 1: would suggest naming the file `useMuxRepay.ts`

sophialittlejohn (approved):

PR #2620, opened 2025-02-08
ci: Review GHA pipeline

gpmayorga:
### Description
This pull request...

- Upgrades all available external GH Actions to newer versions.
- Upgrades the node version of Gcloud functions from v18 to v20 as it deprecates in April
- Configures dependabot to catch new GHA version packages monthly (needs activation in setitngs of the repository)
- Fixes a bug where production deployments might fetch from the wrong artifacts
- Fixes small typo, linting and syntax issues in the workflow files
- Removes catalyst deployment
- Removes deployment of functions for environments: fast-forward (main) & development.
- Add new OWNERS to make sure we have at least 3 as at least 2 are needed to approve a PR (the one creating the change and a reviewer)
- Changes how prod is deployed which now will:
  - Fetch the artifacts form the relese assets
  - Allow to specify a tag that has been released in the past to roll back prod.
  
Screenshot 2025-02-10 at 20 14 49

### Impact

Lots of small changes that should not change the pipline behavior but critical changes were made to:
- Prod deployment, needs testing to make sure it works as intended (needs to be merged first)
- Function deployment masked under NEVER_USED which was causing an actual NEVER_USED function to be deployed in Google repeatedly, causing a lot of worflows to fail (look at main branch commits mostly all red)

### Pending testing

- Test the release workflow for production (needs merging to main first)
- [ ] Test the pinning and onboarding functions (upgraded to node 20)


.github/workflows/e2e.yml
```diff
@@ -25,22 +26,21 @@ jobs:
...
-        env: # env available for only this steps
-          CYPRESS_PRIVATE_KEY: ${{secrets.CYPRESS_PRIVATE_KEY}}
-          PRIVATE_KEY: ${{secrets.CYPRESS_PRIVATE_KEY}}
-          NETWORK_NAME: centrifuge
-          RPC_URL: https://fullnode-apps.demo.k-f.dev
-          CHAIN_ID: 2090
-          SYMBOL: DEVEL
-          IS_TESTNET: true
-          DEBUG: true
+      - name: Create .env file
```
sophialittlejohn, line 29: we're not actually using cypress anymore. Maybe this is a good opportunity to remove it from the workflow completely?

.github/workflows/manual-deploys.yml
```diff
@@ -1,17 +0,0 @@
-name: Deploy to Catalyst and demo
-on:
-  workflow_dispatch:
-      inputs:
-          environment:
-            type: choice
-            description: where to deploy
-            options:
-                - demo
```
sophialittlejohn, line 9: if demo is not triggered manually anymore, how is it deployed?

.github/workflows/prod-deploy.yml
```diff
@@ -4,6 +4,11 @@ on:
     types:
       - released
   workflow_dispatch:
+    inputs:
+      tag:
+        description: 'The release tag to promote (e.g. v1.2.3)'
+        required: true
+        type: string
```
sophialittlejohn, line 11: does this mean we can promote through the actions tab instead of clicking into the release and switching the checkbox from pre-release to release?

.github/workflows/e2e.yml
gpmayorga: Yes but in that case shuold I not also delete:
- The test files under centrifuge-app/tests/e2e and centrifuge-app/synpress.config.ts
- The lines in package.json referencing e2e tests.
- The workflow file

.github/workflows/prod-deploy.yml
gpmayorga: Modified the prod deployment to now retrieve the assets from an existing release, allowing for our current flow but also to roll back to a known release tag manually triggering the job

.github/workflows/e2e.yml
sophialittlejohn: I hope to use Cypress again the same way after we integrate the SDK, so maybe we can just disable the workflow for now?

gpmayorga: done, I've disabled the automatic triggers so this will only run if we click on it

sophialittlejohn (approved):

PR #2643, opened 2025-02-19
ci: Fix functions deploy variables for prod

gpmayorga:
### Description

The deployment of functions for prod was failing:
https://github.com/centrifuge/apps/actions/runs/13414045735

This is due to how the vars.PINNING_API_FUNCT_SA and vars.ONBOARDING_FUNCT_SA is loaded when they are set in the matrix step.
Moved them to be inside the `steps` section where the variables for   `environment: production` are actually set right
Also added some debug to prove that it works withiout deploying the function when in a PR. It is difficult to test production jobs as they would generally only trigger when we want to fully release so I added extra debug stuff to it

#<issue-number>

### Approvals

- [ ] Dev


gpmayorga:
It seems like now it is fetching the right Service Account:
Screenshot 2025-02-19 at 15 40 38

sophialittlejohn (approved):

Issue #2547, opened 2024-11-28
Portfolio Page: Redesign


0x4Graham:
The following ticket is for the Portfolio page redesign. 

![Image](https://github.com/user-attachments/assets/e44ba25f-385a-438b-9fa3-b99f974c6c35)

https://www.figma.com/design/ng7qdNcSCXSDA6ZUdWIs6u/Pool-Overview%2F-Pool-Detail?node-id=4229-17539&node-type=canvas&t=Ws41xCHvJFgmwn8k-0

Starting with the top bar, where we see the 4 following items:
![Image](https://github.com/user-attachments/assets/a10d5c55-169d-478f-afc2-0267664aea90)

- Portfolio Value (Value of the entire portfolio)
- % since inception --> NB: = (Current Portfolio Value + realized & unrealized gains)/investments-redemptions - 1 x 100
- realized profit
- unrealized profit

The banner can reused for the prime portfolio page. 

**Portfolio Performance Block**

![Image](https://github.com/user-attachments/assets/f15e0147-7f7c-4457-8ac1-5e1132a27469)

On the graph, we will show two lines.
1. Total Portfolio Value
2. Total PnL (realized profit + unrealized profit)

**Available to Invest**
![Image](https://github.com/user-attachments/assets/ba8758d7-0f2b-438e-a4b5-4fbf66b2c208)

We need to query the wallet balance of the current chain USDC balance. Just like we currently do on the Invest flow showing available balance. **Right now only add USDC.**

No balance UX
![Image](https://github.com/user-attachments/assets/27acd7f6-5b55-405f-99eb-cc8d9fe3ca6d)

**Investment Positions**

A table including the investment positions per chain:

![Image](https://github.com/user-attachments/assets/fd0028b2-6bd6-46f6-a75a-57fce37854f6)

CTA's open the already built Invest and Redeem side bar. 

The : opens a drop down with the following two CTA's
1. Send
2. Receive 

These will open the send/receive side bars we already have (I think) 

**Investment History**
Table showing the investment history of the portfolio. 

View all should navigate to:

sonam-jo:
Made some edits directly in the ticket above

PR #2652, opened 2025-03-04
New portfolio page

kattylucy:
Add new design to portfolio page 

#2547

### DIFF
--- a/centrifuge-app/src/pages/Portfolio/index.tsx
+++ b/centrifuge-app/src/pages/Portfolio/index.tsx
@@ -169,7 +169,7 @@ function PortfolioDetails({ address, chainId }: { address: string; chainId: numb
 
   return (
     <>
-      <Box borderBottom={`1px solid ${theme.colors.borderPrimary}`} pb={1} mx={2} />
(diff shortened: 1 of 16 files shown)


centrifuge-app/src/pages/Portfolio/index.tsx
```diff
@@ -21,7 +21,8 @@ export default function PortfolioPage() {
 
 function Portfolio() {
   const theme = useTheme()
-  const address = useAddress()
+  const test = useAddress()
+  const address = '0x0d9e269ecc319bae886dd8c8b98f7b89269c2b1b000000000000a4ec45564d00'
   const { showNetworks, connectedNetwork, evm } = useWallet()
```
kattylucy: This is so product have some data to test the portfolio page - will delete de hardcoded address once product has approved

onnovisser (approved):

BrianCarter-Design:
Screenshot 2025-03-19 at 10 24 48
On a less significant note regarding spacing. The alignment on the sections below the header are not aligned to the left and the right of the header section. And the spacing between each data value in the header should be increased so that users can easily see which value is associated with which heading.

Issue #2651, opened 2025-02-27
CFG Migration: Legacy CFG to new CFG


0x4Graham:

0x4Graham:
<img width="548" alt="Image" src="https://github.com/user-attachments/assets/9e56cbff-1ad1-4a6d-b84a-52d417ba21b8" />
On the Cent Chain migration, it shouldn't say WCFG. Should be Legacy CFG

0x4Graham:
Let's add a "Back to portfolio" button here

<img width="701" alt="Image" src="https://github.com/user-attachments/assets/5eba45e3-88b2-48a1-b90a-73be05306f15" />

0x4Graham:
<img width="1296" alt="Image" src="https://github.com/user-attachments/assets/5d8abf58-8aa9-4da3-b65d-e178e16638f1" />

1. Please add Migrate button next to the send and receive for legacy CFG.

2. When connected with an EVM wallet that is connected to Centrifuge chain, the migrate banner doesn't show even though there is a Devrel balance

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/5c8313f0-f9be-4181-b965-653ebbeb8fb0" />

0x4Graham:
<img width="363" alt="Image" src="https://github.com/user-attachments/assets/9a6fbf97-a4c0-4518-b9d1-0022c4ccb038" />

<img width="346" alt="Image" src="https://github.com/user-attachments/assets/fbfa0430-8ab5-458b-9d53-f0dcc7ca22a3" />

Toasts still aren't updated either.

0x4Graham:
It shouldn't be called "newCFG" and the logo/chain should be Ethereum 

<img width="1152" alt="Image" src="https://github.com/user-attachments/assets/f2c008ba-3b16-4f32-ae41-755a72e459c1" />

@kattylucy

0x4Graham:
Transaction Completed, but the UI did not update: 

<img width="1435" alt="Image" src="https://github.com/user-attachments/assets/ae9b2cfa-495e-478a-9480-37591721ac22" />

<img width="788" alt="Image" src="https://github.com/user-attachments/assets/c5505e54-9eeb-415b-9f9e-8930521ed0cf" />

https://testnet.axelarscan.io/gmp/0x29f9d9d54bd399f40c74d3a56b461043db556b0d9bb52360116c9eafad5fe434

kattylucy:
[...]

Issue #2650, opened 2025-02-27
CFG Migration: WCFG -> CFG Flow


0x4Graham:

kattylucy:
FOR TESTING

1) Turn on the debug flag `showCFGTokenMigration`
2) Go to the portfolio page
3) Add the after /portofolio -> /portolio/migrate/eht

0x4Graham:
<img width="380" alt="Image" src="https://github.com/user-attachments/assets/d09283a3-448b-498b-ae58-045e05b5da61" />

Needs to change to "Approve WCFG for Migration"

<img width="368" alt="Image" src="https://github.com/user-attachments/assets/0852865c-c73c-40d5-a300-87df9d6100b3" />

"Migrate WCFG for CFG"

0x4Graham:
@kattylucy Also going to extend the designs for the WCFG to CFG migration. 

Can we show the steps? Approve WCFG and then the actual migration

<img width="511" alt="Image" src="https://github.com/user-attachments/assets/60deee48-2901-4109-abfb-36faedb2dc02" />

https://www.figma.com/design/ng7qdNcSCXSDA6ZUdWIs6u/Pool-Overview%2F-Pool-Detail?node-id=6108-1932&t=IcVzbFqzlw4fpKZY-1

0x4Graham:
@kattylucy the fees are incorrect here and SHOULD NOT come off the CFG token balance. 

<img width="657" alt="Image" src="https://github.com/user-attachments/assets/33a0761f-6926-4df8-bd0f-3066c0a6b8ee" />

I.e here it should still be 100 WCFG. You do not substract the fee from the CFG being sent. It's rather coming from your ethereum balance.

Issue #2649, opened 2025-02-27
CFG Migration: Portfolio Page Update


0x4Graham:
*Legacy CFG = Current CFG on Centrifuge Chain

To support the WCFG migration process, we need to enhance the portfolio page to guide users holding legacy or WCFG tokens through the migration flow. This ticket involves adding a banner and updating the positions table button for these tokens, ensuring both elements navigate to the WCFG migration flow as depicted in the attached images.

**Banner Display**
**Condition**: On the portfolio page, check if the user holds any lefacy or WCFG tokens.
**Action**: If the condition is met, display a banner at the top of the portfolio section.
**Content**: The banner should display the following text:
**Behavior**: Clicking the "Migrate" button should navigate the user to the WCFG/Legacy CFG Flow migration flow.

_Start your CFG migration — click here to begin the process and follow the easy steps. Migrate tokens._

**Positions Table Button**
**Condition**: In the positions table, identify rows where the token is legacy or WCFG.
**Action**: Replace the default button label ("Invest" or "Redeem") with "Migrate" for these tokens.
**Styling**: The "Migrate" button should be yellow to stand out (reference Image 0).
**Behavior**: Clicking the "Migrate" button should navigate the user to the WCFG/Legacy CFG Flow migration flow.

<img width="710" alt="Image" src="https://github.com/user-attachments/assets/ba7895e9-3420-4937-8e0c-0d878406f47e" />

PR #2663, opened 2025-03-17
CFG token migration

kattylucy:
- Flow to migrate from eth to eth 
- Flow to migrate from centrifuge chain to eth
- Banner on portfolio page to redirect to migration token page

#2650 
#2649 
#2651


centrifuge-app/src/pages/Portfolio/useAxelarStatus.ts
```diff
@@ -0,0 +1,34 @@
+import { AxelarGMPRecoveryAPI, Environment } from '@axelar-network/axelarjs-sdk'
```
onnovisser, line 1: This package is 4MB (!!!) minified. Could we lazy-load it?

centrifuge-app/src/pages/Portfolio/useTokenBalance.tsx
```diff
@@ -0,0 +1,52 @@
...
+      new: 'TODO',
+    }
+
+const ABI = ['function balanceOf(address owner) view returns (uint256)']
+
+export const useTokenBalance = (userAddress: string | undefined) => {
+  return useQuery(
+    ['tokenBalance', userAddress],
+    async () => {
+      const provider = new BrowserProvider(window.ethereum)
```
onnovisser, line 25: this should ideally use multicall

```
mport {
  CurrencyBalance,
  evmMulticall,
  EvmMulticallCall,
} from '@centrifuge/centrifuge-js'

export function useCfgBalances(userAddress?: string) {
  const {
    evm: { getProvider },
  } = useWallet()
  return useQuery(
    ['tokenBalance', userAddress],
    async () => {
      const calls: EvmMulticallCall[] = [
        {
          target: cfgConfig.legacy,
          call: ['function balanceOf(address) view returns (uint256)', userAddress!],
          returns: [['legacy']],
          allowFailure: true,
        },
        {
          target: cfgConfig.new,
          call: ['function balanceOf(address) view returns (uint256)', userAddress!],
          returns: [['new']],
          allowFailure: true,
        },
      ]
      const multicallData = await evmMulticall<{ legacy?: bigint; new?: bigint }>(calls, {
        rpcProvider: getProvider(ethConfig.chainId),
      })

      return {
        legacy: {
          balance: new CurrencyBalance(multicallData.legacy ?? '0', 18).toDecimal(),
          currency: currencies.wCFG,
        },
        new: {
          balance: new CurrencyBalance(multicallData.new ?? '0', 18).toDecimal(),
          currency: currencies.newCFG,
        },
      }
    },
    {
      enabled: !!userAddress,
    }
  )
}
```

mustermeiszer:
> A follow-up Ethereum transaction is required to claim your CFG tokens.

Can we remove that, that is not the case.

mustermeiszer:
UI does not store ongoing migrations, right? @kattylucy the indexer will hopefully pick this up once [this](https://github.com/centrifuge/api/issues/281) is done.

centrifuge-app/src/pages/Portfolio/CFGTokenMigrationCent.tsx
```diff
@@ -0,0 +1,373 @@
...
+          setAxelarHash(axelarHash)
+          setStep(2)
+          const transactionData: TransactionData = {
+            from_address: address,
+            to_address: evmAddress,
+            tx_hash: result.txHash,
+            chain: 'centrifuge',
+            amount: balance?.toNumber(),
+          }
+          await recordTransaction(transactionData)
```
mustermeiszer, line 127: @kattylucy what kind of event are we filtering for and what kind of hash are we looking for? The main issue is, that the we queue before submitting to Axelar, which in case of congestion or multiple parallel transactions COULD lead to some different behaviour. 

I would suggest you and Graham try to submit as many parallel migrations as possible to see whether the UI actually works.

mustermeiszer: My main worry is, that it will be hard to index that once we have more than one migrationb per block. But for now its working great. The Indexer will need to separate those cleanly

